PR #1, opened 2021-12-07
Initial Implementation

cretz:
This is the initial implementation. Notes:

* TypeScript is not plugged into the runner due to late decision on how to handle versions, so it will be added as the next PR.
* It may be easier to browse code at https://github.com/cretz/temporal-sdk-features/tree/initial-impl (as of the time of this PR, branch may be deleted on merge of course). At the very least, the README there is much easier to read than this PR.
* Discussion still needs to occur on what a CI run of this looks like both inside a repository (i.e. using the latest commit of that repository) and in this repo and/or periodic runs.


cretz:
Closing by request and making several different PRs :-( Unfortunately there's so much boilerplate that some are finding it hard to read (there's only like 20 real files to read).

PR #5, opened 2021-12-14
Activity start race bug attempt, Go SDK update, other minor features

cretz:
## What was changed

* Added (unsuccessful) attempt to replicate a user bug at https://github.com/temporalio/sdk-go/issues/670
* Updated Go SDK to v1.12.0
* Added support for local-path Go SDK version

## Why?

Was hard to get the event order described. Committing this because it may have value anyways even if it didn't replicate a bug.


features/bugs/go/activity_start_race/feature.go
```diff
@@ -0,0 +1,138 @@
...
+	pollWorkflowWait.Done()
+
+	// Wait a bit, then cancel the workflow
+	time.Sleep(activitySleep * 2)
+	r.Require.NoError(r.Client.CancelWorkflow(ctx, run.GetID(), run.GetRunID()))
+	return run, err
+}
+
+func CheckHistory(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
+	// We want to do the default history check which just does a replay of the
```
bergundy, line 70: Can't we make this completely deterministic?
And if you're using manually generated history, do we need this complexity in the feature? Are you just keeping it around for future use-cases where we need to try and reproduce timing sensitive histories?

bergundy (approved):

features/bugs/go/activity_start_race/feature.go
cretz: > Can't we make this completely deterministic?

I cannot make it completely deterministic I don't think without slowing things way down. My attempts to make it reliably return the same order of events failed, I can only get it to do so _sometimes_.

>  And if you're using manually generated history, do we need this complexity in the feature? Are you just keeping it around for future use-cases where we need to try and reproduce timing sensitive histories?

I actually generated this history, it just doesn't happen in this order every time. I keep it around for this specific replay. So yeah, there's kinda two separate tests here: 1) that the workflow runs as expected and 2) that the workflow works with the given history. The history of #1 is not checked/used.

PR #14, opened 2022-01-25
Typescript Runner

Sushisource:
Haven't added a replayer yet but this lays down the foundation for a runner. Works with local and nonlocal versions.


Sushisource:
Will also likely add some CI in the next PR

gradlew
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env sh
```
bergundy, line 1: Should this be committed?

harness/ts/main.ts
```diff
@@ -38,14 +47,16 @@ async function run() {
   for (const featureAndTaskQueue of opts.featureAndTaskQueues) {
     // Split
     const colon = featureAndTaskQueue.indexOf(':');
-    const featureDir = featureAndTaskQueue.substring(0, colon);
+    const cpos = colon == -1 ? undefined : colon;
```
bergundy, line 50:
```suggestion
    const cpos = colon === -1 ? undefined : colon;
```

bergundy: This is more readable imo (assuming single colon for simplicity):
```ts
const [featureDir, taskQueueFromOpt] = featureAndTaskQueue.split(':');
const taskQueue = taskQueueFromOpt ?? featureDir;
```

harness/ts/main.ts
```diff
@@ -38,14 +47,16 @@ async function run() {
...
+    const featureDir = featureAndTaskQueue.substring(0, cpos);
     const taskQueue = featureAndTaskQueue.substring(colon + 1);
 
     let runner;
     try {
       // Find the source
       const source = sources.find((s) => s.relDir === featureDir);
       if (!source) {
+        // noinspection ExceptionCaughtLocallyJS
```
bergundy, line 59: What's inspecting this?

harness/ts/package.json.tmpl
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "dependencies": {
+  {{ if .LocalSDK }}
+    "@temporalio/activity": "{{ .LocalSDK }}/packages/activity",
+    "@temporalio/client": "{{ .LocalSDK }}/packages/client",
+    "@temporalio/common": "{{ .LocalSDK }}/packages/common",
+    "@temporalio/proto": "{{ .LocalSDK }}/packages/proto",
+    "@temporalio/worker": "{{ .LocalSDK }}/packages/worker",
+    "@temporalio/workflow": "{{ .LocalSDK }}/packages/workflow",
+    "temporalio": "{{ .LocalSDK }}/packages/meta",
```
bergundy, line 15: You probably don't even need to install this

harness/ts/package.json.tmpl
```diff
@@ -0,0 +1,30 @@
+{
```
bergundy, line 1: Don't love generating JSON (or other forms of structured data) from string templates but ok..

harness/ts/tsconfig.json.tmpl
```diff
@@ -0,0 +1,20 @@
...
+  "extends": "@tsconfig/node16/tsconfig.json",
+  "version": "4.4.2",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "./tslib",
+    "rootDirs": ["../", "."],
+    "paths": {
+      "@temporalio/harness": ["../harness/ts/harness.ts"],
+
+      "*": ["node_modules/*", "node_modules/@types/*"]
```
bergundy, line 11: Interesting that all of this is needed

harness/ts/tsconfig.json.tmpl
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "typeRoots": ["node_modules/@types"],
```
bergundy, line 13: Why do we need this?

bergundy (approved):

cmd/run_typescript.go
```diff
@@ -0,0 +1,123 @@
...
+		MetaPkgVersion: MetaPkgVersion,
+	})
+	if err != nil {
+		return fmt.Errorf("failed build package.json template: %w", err)
+	}
+	pkgJsonDest, err := os.Create(filepath.Join(tempDir, "package.json"))
+	if err != nil {
+		return fmt.Errorf("failed create package.json in harness: %w", err)
+	}
+	_, err = pkgJsonDest.WriteString(packageJsonEvaluated.String())
```
cretz, line 79: `os.WriteFile(filepath.Join(tempDir, "package.json"), packageJSONEvaluated.Bytes(), 0644)` is easier (same for other file writes)

cmd/run_typescript.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to make absolute path for TS harness: %w", err)
+	}
+	packageJson, err := template.ParseFiles(filepath.Join(harnessPath, "package.json.tmpl"))
+	if err != nil {
+		return fmt.Errorf("failed to load package.json template: %w", err)
+	}
+
+	// Create package.json from template
+	packageJsonEvaluated := bytes.NewBufferString("")
```
cretz, line 59: Often clearer as `var packageJsonEvaluated bytes.Buffer` (and use a pointer to `Execute` below)

cmd/run_typescript.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return fmt.Errorf("failed create package.json in harness: %w", err)
+	}
+	_, err = pkgJsonDest.WriteString(packageJsonEvaluated.String())
+	if err != nil {
+		return fmt.Errorf("failed to write to package.json in harness: %w", err)
+	}
+
+	// Copy tsconfig
+	tsConfigSrc, err := os.Open(filepath.Join(harnessPath, "tsconfig.json.tmpl"))
```
cretz, line 85: You don't close the files you open. But if you want, for small files, doing `os.ReadFile` and `os.WriteFile` is easier and memory isn't really a concern.

gradlew
cretz: Looks like different gradle versions spit out different forms of this file. But I do think this should be committed because I didn't have `+x` on my original (forgot to do it, was on Windows, derp).

harness/ts/main.ts
Sushisource: IntelliJ. I actually added it by accident but, it hides the warning which is nice I suppose? :man_shrugging:

harness/ts/tsconfig.json.tmpl
Sushisource: When using local linking NPM doesn't actually install transitive dependencies properly for some crazy reason, and then on top of that because of way I have to point to a different dir here it just doesn't search for `node_modules` in the right place, and even when you specify just that, it still fails to find types properly. :man_shrugging:  node.

harness/ts/package.json.tmpl
Sushisource: Yeah I suppose that makes perfect sense in retrospect but this was easy to do with built-in Go stdlib stuff

cmd/run_typescript.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return fmt.Errorf("failed creating temp dir: %w", err)
+	}
+	r.log.Info("Building temporary Typescript project", tag.NewStringTag("Path", tempDir))
+
+	// Remove when done if configured to do so
+	// TODO: Dedupe
+	if !r.config.RetainTempDir {
+		c := make(chan os.Signal)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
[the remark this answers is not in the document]
Sushisource: It's not - the temp dir is not cleaned if you hit ctrl-c. I was going to apply this to all the runners

cmd/run_typescript.go
Sushisource: Bah! No close-on-drop! Thanks for all the bits of advice

gradlew
Sushisource: It's just a slightly newer version. In general though @bergundy gradle init scripts are always committed

cmd/run_typescript.go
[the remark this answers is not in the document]
cretz: > the temp dir is not cleaned if you hit ctrl-c.

We can do it at the top level and have it close the context and count on each piece handle the context. But what I was saying here is just move the signal notifier to the function itself, not only for temp dir deletion. It should stop this function too.

cretz (approved):

PR #24, opened 2022-02-25
Don't use nodeModulesPaths for TS harness

bergundy:
This WorkerOption was removed in a recent TS SDK PR.

### DIFF
--- a/.github/workflows/typescript.yaml
+++ b/.github/workflows/typescript.yaml
@@ -44,7 +44,6 @@ jobs:
         with:
           go-version: '^1.17'
 
-
       # Don't build during install phase since we're going to explicitly build
       - run: npm ci --ignore-scripts
       - run: npm run build
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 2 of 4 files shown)


lorensr:
Ran this twice, both times ended with:

image

[build-ts](https://github.com/temporalio/sdk-features/runs/5339766509?check_suite_focus=true#logs)

.github/workflows/typescript.yaml
```diff
@@ -60,6 +64,7 @@ jobs:
         run: docker-compose up -d
       - name: Wait for Temporal Server
         run: npm run wait-namespace
+        working-directory: ./sdk-ts
```
Sushisource, line 67: This isn't needed / isn't the problem since it's the default working dir specified above.

Maybe we need to specify `main` since it defaults to `master`? Not sure why this would've changed though.

Sushisource: Oh I see what's going on here:
image

Some how the stupid checkout action is checking out the features repo when it should be checking out the ts repo and it's completely unclear why that is.

See @bergundy I told you we should just dupe this in both repos lol

bergundy: I got it already, just haven't found a fix using the DSL

Sushisource:
After this is fixed we can merge https://github.com/temporalio/sdk-typescript/pull/502 as well

Sushisource (approved):

PR #64, opened 2022-06-27
Python Support

cretz:
## What was changed

Added Python harness, initial Python feature, and a CI test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,6 +39,23 @@ jobs:
       - run: npm run build
       - run: npm run lint
 
+  build-python:
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest] #  windows-latest - like 8x slower. Excluded for now since we're just building.
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Print build information
+        run: 'echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}, os: ${{ matrix.os }}'
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.10'
+      - run: python -m pip install --upgrade wheel poetry poethepoet
+      - run: poetry install --no-root
+      - run: poe lint
+
   feature-tests-ts:
     # GH actions is a bit limited here and won't accept the current ref.
     # When changing typescript.yaml, make sure to update the branch name below.
@@ -54,6 +71,14 @@ jobs:
     with:
       go-repo-path: temporalio/sdk-go
       go-repo-ref: master
+
+  feature-tests-python:
+    # GH actions is a bit limited here and won't accept the current ref.
+    # When changing python.yaml, make sure to update the branch name below.
+    uses: cretz/temporal-sdk-features/.github/workflows/python.yaml@python
+    with:
+      python-repo-path: temporalio/sdk-python
+      python-repo-ref: main
 
   build-java:
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,16 +65,12 @@ jobs:
       typescript-repo-ref: main
 
   feature-tests-go:
-    # GH actions is a bit limited here and won't accept the current ref.
-    # When changing go.yaml, make sure to update the branch name below.
-    uses: temporalio/sdk-features/.github/workflows/go.yaml@main
+    uses: ./.github/workflows/go.yaml
     with:
       go-repo-path: temporalio/sdk-go
       go-repo-ref: master
 
   feature-tests-python:
-    # GH actions is a bit limited here and won't accept the current ref.
-    # When changing python.yaml, make sure to update the branch name below.
     uses: ./.github/workflows/python.yaml
     with:
       python-repo-path: temporalio/sdk-python
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,7 +75,7 @@ jobs:
   feature-tests-python:
     # GH actions is a bit limited here and won't accept the current ref.
     # When changing python.yaml, make sure to update the branch name below.
-    uses: cretz/temporal-sdk-features/.github/workflows/python.yaml@python
+    uses: ./.github/workflows/python.yaml
     with:
       python-repo-path: temporalio/sdk-python
       python-repo-ref: main
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -1,0 +1,64 @@
+name: Python Features Testing
+on:
+  workflow_call:
+    inputs:
+      python-repo-path:
+        required: true
+        type: string
+        default: 'temporalio/sdk-python'
+      python-repo-ref:
+        required: true
+        type: string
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,6 +72,14 @@ jobs:
       go-repo-path: temporalio/sdk-go
       go-repo-ref: master
 
+  feature-tests-python:
+    # GH actions is a bit limited here and won't accept the current ref.
+    # When changing python.yaml, make sure to update the branch name below.
+    uses: cretz/temporal-sdk-features/.github/workflows/python.yaml@python
```
cretz, line 78: This will be changed before merge

harness/python/feature.py
```diff
@@ -0,0 +1,120 @@
...
+    expect_activity_error: Optional[str] = None,
+    file: Optional[str] = None,
+    start: Optional[Callable[[Runner], Awaitable[WorkflowHandle]]] = None,
+    check_result: Optional[Callable[[Runner, WorkflowHandle], Awaitable[None]]] = None,
+) -> None:
+    if not file:
+        file = inspect.stack()[1].filename
+    # Split the file path to get the last two dirs if present
+    parts = Path(file).parts
+    assert len(parts) >= 3, "Expected at least three parts to the file path"
```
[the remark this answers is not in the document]
cretz: > Worth printing the full path here.

This isn't an expected error. In retrospect I shouldn't have even put a message on the assertion. But ok, now I will remove the assertion and make this a normal error.

> Maybe just have caller explicitly pass it in.

In Go we derive it from the function pointer, in Java from the package name, in TS from the crawled path, and here from the caller file. None of them really have the caller pass it in. All of them are a bit fragile in their own way.

.github/workflows/python.yaml
```diff
@@ -0,0 +1,64 @@
...
+          submodules: recursive
+          path: sdk-python
+          ref: ${{ inputs.python-repo-ref }}
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.17'
+
+      # Build SDK
+      - uses: Swatinem/rust-cache@v1
```
bergundy, line 45: Doesn't this need to be pointed at the right directory? https://github.com/temporalio/sdk-python/blob/25caaf008f50cd446b9c7791ea97f3af3bfe8717/.github/workflows/ci.yml#L33

cmd/run.go
```diff
@@ -51,13 +51,13 @@ func (r *RunConfig) flags() []cli.Flag {
...
 			Name:        "lang",
-			Usage:       "SDK language to run ('go' or 'java' or 'ts')",
+			Usage:       "SDK language to run ('go' or 'java' or 'ts' or 'py')",
 			Required:    true,
 			Destination: &r.Lang,
 		},
 		&cli.StringFlag{
 			Name:        "version",
-			Usage:       "SDK language version to run. Typescript versions may start with `/` to use a local SDK via an absolute path.",
+			Usage:       "SDK language version to run. Most languages support versions as paths.",
```
bergundy, line 60: Does path have to be absolute?

cmd/run_python.go
```diff
@@ -0,0 +1,83 @@
...
+			// We expect a dist/ directory with a single whl file present
+			wheels, err := filepath.Glob(filepath.Join(r.config.Version, "dist/*.whl"))
+			if err != nil {
+				return fmt.Errorf("failed glob wheel lookup: %w", err)
+			} else if len(wheels) != 1 {
+				return fmt.Errorf("expected single dist wheel, found %v", len(wheels))
+			}
+			versionStr = "{ path = " + strconv.Quote(wheels[0]) + " }"
+		}
+		pyProjectTOML := `
```
bergundy, line 44: Wouldn't you rather generate this with a structure that maps to toml? E.g: https://github.com/pelletier/go-toml

bergundy (approved):
Nothing blocking

harness/python/feature.py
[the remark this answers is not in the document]
cretz: Changed to error

.github/workflows/ci.yaml
cretz: Changed to relative

.github/workflows/ci.yaml
```diff
@@ -55,6 +72,14 @@ jobs:
       go-repo-path: temporalio/sdk-go
       go-repo-ref: master
 
+  feature-tests-python:
+    # GH actions is a bit limited here and won't accept the current ref.
+    # When changing python.yaml, make sure to update the branch name below.
+    uses: ./.github/workflows/python.yaml
```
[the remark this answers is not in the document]
cretz: Yeah, just noticed this when reading https://docs.github.com/en/actions/using-workflows/reusing-workflows. Will remove the comment.

.github/workflows/python.yaml
cretz: Ah, it does indeed, will fix.

cmd/run.go
cretz: Hrmm, I'll have to check. For Python I do a simple glob which would work for cwd but I don't seem to make it absolute when putting in the pyproject.toml. I will confirm.

cmd/run_python.go
cretz: Nah, doesn't seem necessary for such small use. This is a single, small temp file. Were I to do more advanced TOML mods, definitely.

harness/python/main.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    args = parser.parse_args()
+
+    # Configure logging
+    logging.basicConfig(level=getattr(logging, args.log_level.upper()))
+
+    # Collect all feature paths
+    root_dir = Path(__file__, "../../../features").resolve()
+    rel_dirs = sorted(
+        [
```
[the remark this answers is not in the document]
cretz: Yeah, I was creating a list comprehension but I can switch it to a generator expression

.github/workflows/ci.yaml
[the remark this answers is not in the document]
cretz: Fixed and also fixed Go to use this too (will leave TS for y'all, didn't want to break something I was unsure about)

.github/workflows/python.yaml
cretz: Fixed

cmd/run.go
cretz: I was not accepting relative paths, but I do now. There is a weird bug with Poetry where an absolute path on Windows had to have the drive letter capitalized, so I did that too.

harness/python/main.py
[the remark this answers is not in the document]
cretz: Fixed

harness/python/main.py
```diff
@@ -0,0 +1,68 @@
...
+    parser.add_argument(
+        "features", help="Features as dir + ':' + task queue", nargs="+"
+    )
+    args = parser.parse_args()
+
+    # Configure logging
+    logging.basicConfig(level=getattr(logging, args.log_level.upper()))
+
+    # Collect all feature paths
+    root_dir = Path(__file__, "../../../features").resolve()
```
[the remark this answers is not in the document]
cretz: :shrug:

Issue #77, opened 2022-07-20
[Feature Request] Implement activity features


bergundy:

cretz:
This has already found two bugs:

* https://github.com/temporalio/sdk-go/issues/860
* https://github.com/temporalio/sdk-go/issues/859

PR #85, opened 2022-07-21
Activity try cancel feature

cretz:
## What was changed

Add activity try-cancel feature.

## Checklist

1. Part of #77

I'll get to the rest of the features later


features/activity/cancel_try_cancel/README.md
```diff
@@ -7,10 +7,19 @@ using the Try Cancel mode. The implementation should demonstrate that the activi
...
+## Detailed spec
+
 * When the SDK issues the activity cancel request command, server will write an
   activity cancel requested event to history
 * The workflow immediately resolves the activity with its result being cancelled
 * Server will notify the activity cancellation has been requested via a response
   to activity heartbeating
-* The activity may ignore the cancellation request if it explicitly chooses to do
\ No newline at end of file
+* The activity may ignore the cancellation request if it explicitly chooses to do
```
bergundy, line 17:
```suggestion
* The activity may ignore the cancellation request if it explicitly chooses to
```

features/activity/cancel_try_cancel/feature.workflow.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!(e instanceof ActivityFailure) || !(e.cause instanceof CancelledFailure)) {
+      const error = e instanceof Error ? e : new Error(`${e}`);
+      throw new ApplicationFailure('Expected cancellable', null, true, undefined, error);
+    }
+  }
+
+  // Confirm signal is received saying the activity got the cancel
+  let activityResult: string | undefined;
+  setHandler(activityResultSignal, (v) => void (activityResult = v));
+  await condition(() => !!activityResult, 10000);
```
bergundy, line 50: Do we have to set a timeout here?
Wouldn't it be better to implement the test in a naive manner and set the timeout in the runner as a general strategy?

features/activity/cancel_try_cancel/feature.ts
```diff
@@ -0,0 +1,48 @@
...
+  activities: {
+    cancellableActivity,
+  },
+  execute: async (runner) => {
+    client = runner.client;
+    return await runner.executeSingleParameterlessWorkflow();
+  },
+});
+
+let client: WorkflowClient | undefined;
```
bergundy, line 18: Not loving this global variable.
Would have been much nicer to have a way to instantiate an activity class (or closure) or even inject this in an activity interceptor.

features/activity/cancel_try_cancel/feature.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!client) {
+    throw new Error('Missing client');
+  }
+
+  // Heartbeat every second for a minute
+  let result = 'timeout';
+  for (let i = 0; i < 60; i++) {
+    // Wait for a second or until cancelled
+    try {
+      await Promise.race([Context.current().sleep(1000), Context.current().cancelled]);
```
bergundy, line 31: `sleep` is already cancellation aware

bergundy: You're not hearbeating in the loop, how will the activity get cancelled?

features/activity/cancel_try_cancel/feature.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch (e) {
+      // Exit loop if cancelled or rethrow if other error
+      if (e instanceof CancelledFailure) {
+        result = 'cancelled';
+        break;
+      }
+      throw e;
+    }
+    await Context.current().sleep(1000);
```
bergundy, line 40: Why the extra sleep?

features/activity/cancel_abandon/README.md
```diff
@@ -12,4 +12,7 @@ cancelled
 * When the SDK requests cancel of an abandon mode activity, the workflow code
   immediately unblocks, the activity result being cancelled.
 * Nothing is sent to the server, the activity worker never is notified of the
-  cancellation attempt
\ No newline at end of file
+  cancellation attempt
+
+
+**NOTE:** This feature does not exist in the Go SDK
```
Sushisource, line 18: I fully acknowledge this is a bit anal, but maybe add a "as of <date>" next to this line, since in theory everyone should _eventually_ be conformant

features/activity/cancel_try_cancel/feature.java
```diff
@@ -0,0 +1,116 @@
...
+    private CompletablePromise<String> activityResult;
+
+    @Override
+    public void workflow() {
+      activityResult = Workflow.newPromise();
+      var activities = activities(feature.class, builder -> builder
+              .setScheduleToCloseTimeout(Duration.ofMinutes(1))
+              .setHeartbeatTimeout(Duration.ofSeconds(5))
+              // Disable retry
+              .setRetryOptions(RetryOptions.newBuilder().setMaximumAttempts(1).build())
```
Sushisource, line 43: Does that actually matter here? Shouldn't cancel error be non-retryable anyway? (Noticed this applies to all of them)

features/activity/cancel_try_cancel/feature.java
```diff
@@ -0,0 +1,116 @@
...
+      scope.run();
+
+      // Sleep for short time (force task turnover)
+      Workflow.sleep(1);
+
+      // Cancel the activity and confirm it gets cancelled
+      scope.cancel();
+      try {
+        activityPromise.get().get();
+        throw ApplicationFailure.newFailure("No error", "NoError");
```
Sushisource, line 60: Just a touch more clear
```suggestion
        throw ApplicationFailure.newFailure("Activity should have thrown cancellation error", "NoError");
```

features/activity/cancel_try_cancel/feature.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public void activityResult(String result) {
+      activityResult.complete(result);
+    }
+
+    private WorkflowClient client;
+
+    @Override
+    public Run execute(Runner runner) throws Exception {
+      client = runner.client;
```
Sushisource, line 87: This happened in Go and Java, might make sense to make it a thing available by default somehow?

features/activity/cancel_try_cancel/feature.workflow.ts
Sushisource: +1 to that. Would be nice to have default timeout in runner than can be overriden, and just terminates the workflow with a "timed out" reason or something.

Sushisource (approved):
LGTM. All my comments are nits - Roey's bit about the heartbeat seemingly not making sense in the TS workflow is important tho

features/activity/cancel_try_cancel/feature.py
```diff
@@ -0,0 +1,86 @@
...
+    @workflow.run
+    async def run(self) -> None:
+        # Start workflow
+        handle = workflow.start_activity(
+            cancellable_activity,
+            schedule_to_close_timeout=timedelta(minutes=1),
+            heartbeat_timeout=timedelta(seconds=5),
+            # Disable retry
+            retry_policy=RetryPolicy(maximum_attempts=1),
+            cancellation_type=workflow.ActivityCancellationType.TRY_CANCEL,
```
bergundy, line 27: I know we want to use this code for documentation but I think there's value in sharing this kind of configuration between the different implementations

features/activity/cancel_try_cancel/feature.java
```diff
@@ -0,0 +1,116 @@
...
+      Objects.requireNonNull(client);
+
+      // Heartbeat every second for a minute
+      var result = "timeout";
+      for (int i = 0; i < 60; i++) {
+        try {
+          Thread.sleep(1000);
+        } catch (InterruptedException e) {
+          throw Activity.wrap(e);
+        }
```
bergundy, line 102: Can you explain why you need to catch this to someone like me who's not familiar with the Java SDK please?

features/activity/cancel_try_cancel/feature.java
```diff
@@ -0,0 +1,116 @@
...
+@ActivityInterface
+public interface feature extends Feature, SimpleWorkflow {
+  @SignalMethod
+  void activityResult(String result);
+
+  @ActivityMethod
+  void cancellableActivity();
+
+  class Impl implements feature {
+    private CompletablePromise<String> activityResult;
```
bergundy, line 34: Curious why you're not using `Workflow.await` here like in the other langs

features/activity/cancel_try_cancel/feature.go
```diff
@@ -0,0 +1,79 @@
...
+	// Sleep for the smallest amount of time (force task turnover)
+	if err := workflow.Sleep(ctx, 1*time.Millisecond); err != nil {
+		return err
+	}
+
+	// Cancel and confirm the activity errors with the cancel
+	actCancel()
+	if err := fut.Get(ctx, nil); !temporal.IsCanceledError(err) {
+		return harness.AppErrorf("expected activity cancel error, got: %v", err)
+	}
```
bergundy, line 47: No timeout here? There's a timeout in all the language implementations.

features/activity/cancel_try_cancel/feature.py
bergundy: I saw that you have a RetryDisabled constant in Go.
Maybe make the entire activity config for long-running-at-most-once a constant in all harnesses?

features/activity/cancel_try_cancel/feature.workflow.ts
bergundy: nit:
```ts
const activityResult = await new Promise<string>((resolve) => setHandler(activityResultSignal, resolve));
```

Or with timeout:
```ts
const activityResult = await new Promise<string>((resolve, reject) => {
  setHandler(activityResultSignal, resolve);
  sleep(10000).then(() => reject(ApplicationFailure.nonRetryable('timed out')));
});
```

bergundy (approved):
Overall looks good

features/activity/cancel_try_cancel/feature.java
bergundy: Hmm.. yeah, this would make the test flakier.
Not a concern now but eventually it might.

features/activity/cancel_try_cancel/feature.java
bergundy: Yeah, perhaps, add a default activity interceptor in all langs to inject this into the activity context.

features/activity/cancel_abandon/README.md
cretz: Sure, or "currently" or whatever.

features/activity/cancel_try_cancel/feature.go
cretz: I can add one, though it's not used as much in Go as it is in other languages and a test running forever is just as much of a non-success as a test that throws a timeout.

features/activity/cancel_try_cancel/feature.java
cretz: This is basically the same thing. I just found it cleaner to populate this in Java. But I can change it.

features/activity/cancel_try_cancel/feature.java
cretz: In my head "if I don't want an activity to ever retry, I don't let it". Basically I want any failure of the activity for these tests to be considered non-retryable, not just the expected one.

features/activity/cancel_try_cancel/feature.java
cretz: Meh, clarity on a never-should-be-thrown error in a feature is hardly worth it I think. These errors have stack traces. I guess I'll add it for y'all.

features/activity/cancel_try_cancel/feature.java
cretz: I think we should avoid including any non-activity package things by default to a user (and even if we did, a default interceptor means we might as well hardcode it, an interceptor is an unnecessary abstraction). But for the harnesses here as a helper, sure.

IMO, a Temporal client is no more important than a DB client and the pattern to include X client should be the same for all values of X IMO.

features/activity/cancel_try_cancel/feature.java
cretz: It's a checked exception and I don't let this activity throw checked exceptions, so the compiler complains. This turns it to an unchecked exception that can be thrown out of the activity.

features/activity/cancel_try_cancel/feature.py
cretz: As we start to reuse, sure. Let me get to the next set and if that schedule to close and heartbeat and such remain the same, I'll definitely create something shared.

features/activity/cancel_try_cancel/feature.ts
cretz: This feature is not yet released in Python

features/activity/cancel_try_cancel/feature.ts
cretz: :+1: Will Fix

features/activity/cancel_try_cancel/feature.ts
cretz: This is mistaken leftover code, will remove

features/activity/cancel_try_cancel/feature.workflow.ts
cretz: > Do we have to set a timeout here?

No, I just figure it's best practice for anything you expect to happen soon

>  Wouldn't it be better to implement the test in a naive manner and set the timeout in the runner as a general strategy?

Yes. If we don't already I can make that happen.

> nit:

Ok

features/activity/cancel_try_cancel/feature.java
bergundy: I disagree.
I actually think it's super useful in an activity for things like:
- query / signal my own workflow
- perform a checkpoint and wait for it to be committed (bypassing the heartbeat throttler)

features/activity/cancel_try_cancel/feature.java
bergundy: No biggie, just curious

features/activity/cancel_try_cancel/feature.java
cretz: Of course it's useful. So are many other clients. I don't think it's any more useful than other clients though. If it's too difficult for them to get an external thing, be it our client or another client, we need to fix that. Not assume that the Temporal client is special. They have it at worker creation and activity registration time, there's no reason they can't make it available to the activity.

features/activity/cancel_try_cancel/README.md
cretz: Fixed (was from an older PR)

features/activity/cancel_try_cancel/feature.workflow.ts
cretz: Removed timeouts, put 1 min default workflow execution timeouts on all harnesses, changed activity result

features/activity/cancel_try_cancel/feature.ts
cretz: Fixed

features/activity/cancel_abandon/README.md
cretz: Updated to clarify

features/activity/cancel_try_cancel/feature.java
cretz: Updated

features/activity/cancel_try_cancel/feature.ts
cretz: Removed

features/activity/cancel_try_cancel/feature.java
cretz: Updated

features/activity/cancel_try_cancel/feature.go
cretz: Removed timeout on all of them

PR #87, opened 2022-07-22
Support declaring TS workflows in the feature file

bergundy:
It's not super elegant and might make it harder for others to contribute but thought I'd submit and see what others think.


lorensr (approved):
I don't mind 2 files, but also don't think this is bad. Lint warnings

features/activity/retry_on_error/feature.ts
```diff
@@ -1,26 +1,46 @@
...
+  await activities.alwaysFailActivity();
 }
+
+const activitiesImpl = {
+  async alwaysFailActivity() {
+    throw new Error(`activity attempt ${Context.current().info.attempt} failed`);
+  },
+};
+
+export const feature =
```
cretz, line 31: We can change the harness from `new Feature` to a static `Feature.create` or similar that returns a no-op feature in workflow context. This way, you still can have `proxyActivities<typeof feature.activities>`.

features/activity/retry_on_error/feature.ts
```diff
@@ -1,26 +1,46 @@
 import { Context } from '@temporalio/activity';
+import * as wf from '@temporalio/workflow';
```
cretz, line 2: I think this would set the pattern for everyone to follow, so we have to decide if that's what we want. Aliasing to "wf" what we want? Or maybe non-abbreviated "workflow" instead (we can change the workflow function name)? Would we rather import the parts of workflow we need even though we're in a shared file here?

cretz (commented):
I have no strong preference either way. Two files or one works for me.

features/activity/retry_on_error/feature.ts
bergundy: I find this most convenient and have been using this pattern for a while.
It's just my personal preference. I prefer `wf` over `workflow` because it's less typing and makes shorter lines.
Also would rather have all the utilities available without needing to add an import every time I discover something new's missing.

features/activity/retry_on_error/feature.ts
bergundy: You can't use anything imported from the harness in workflow code with this approach.
I tried to make the class workflow safe but encountered some issues with path aliases.
I think we could get rid of the condition eventually but it might require a bit more flexibility from the SDK bundler.

cretz: > You can't use anything imported from the harness in workflow code with this approach.

I don't think you should use any non-static thing from the harness in the workflow (or would have to use globals instead like I had to in another test to access the runner/client from the activity). Utilities make sense, but things that are dependent on the current execution I hope are unnecessary inside the workflow.

bergundy:
Merging since nobody objected.

features/activity/retry_on_error/feature.ts
cretz: > I don't think you should use any non-static thing from the harness in the workflow

I am assuming agreement on this statement since nobody objected :-)

bergundy: Agreed

PR #106, opened 2022-08-05
Command to "prepare" (i.e. prebuild) SDK directories to run

### LINKED ISSUE #104: [Feature Request] Support "preparing" an SDK to run
### Is your feature request related to a problem? Please describe.

Right now the CLI makes a temp dir and has the package manager lazily download things when it's run. We want to bake all pre-run steps into a container.

### Describe the solution you'd like

* Need a command to "prepare" a dir for run the same way we prepare a temp dir during run today
* Need to be able to provide `--prepared-dir` to the `run` command to use a prepared directory we won't delete instead of a temp dir we do delete

Needed by https://github.com/temporalio/sdk-features/issues/71

cretz:
## What was changed

New `prepare` command that lets you prepare a directory before run.

## Why?

Some builds take a long time, this allows one to prebuild once and run against it multiple times.

## Checklist

1. Closes #104

### DIFF
--- a/cmd/prepare.go
+++ b/cmd/prepare.go
@@ -1,0 +1,106 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/urfave/cli/v2"
+	"go.temporal.io/server/common/log"
+)
+
+func prepareCmd() *cli.Command {
+	var config PrepareConfig
+	return &cli.Command{
+		Name:  "prepare",
+		Usage: "prepare an SDK for execution",
+		Flags: config.flags(),
+		Action: func(ctx *cli.Context) error {
+			return NewPreparer(config).Prepare(ctx.Context)
+		},
+	}
+}
+
+// PrepareConfig is configuration for NewPreparer.
+type PrepareConfig struct {
+	Dir     string
+	Lang    string
+	Version string
+}
+
+func (p *PrepareConfig) flags() []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{
+			Name: "dir",
+			Usage: "Directory name to prepare SDK for run at. " +
+				"This will be relative to the SDK features directory and cannot exist yet.",
+			Required:    true,
+			Destination: &p.Dir,
+		},
+		&cli.StringFlag{
+			Name:        "lang",
+			Usage:       "SDK language to run ('go' or 'java' or 'ts' or 'py')",
+			Required:    true,
+			Destination: &p.Lang,
+		},
+		&cli.StringFlag{
+			Name:        "version",
+			Usage:       "SDK language version to run. Most languages support versions as paths.",
+			Required:    true,
+			Destination: &p.Version,
+		},
+	}
+}
+
+type Preparer struct {
+	log     log.Logger
+	config  PrepareConfig
+	rootDir string
+}
+
+func NewPreparer(config PrepareConfig) *Preparer {
+	return &Preparer{
+		// TODO(cretz): Configurable logger
+		log:     log.NewCLILogger(),
+		config:  config,
+		rootDir: rootDir(),
+	}
+}
+
+func (p *Preparer) Prepare(ctx context.Context) error {
+	var err error
+	if p.config.Lang, err = normalizeLangName(p.config.Lang); err != nil {
+		return err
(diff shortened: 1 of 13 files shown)


cmd/prepare.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/urfave/cli/v2"
+	"go.temporal.io/server/common/log"
+)
+
+func prepareCmd() *cli.Command {
+	var config PrepareConfig
+	return &cli.Command{
+		Name:  "prepare",
+		Usage: "prepare an SDK for execution",
```
bergundy, line 18: (to be used with --prepare-dir) or something meaningful that explains what it means to prepare please

cretz: I put that info in the README. I think this should remain very brief.

cmd/run.go
```diff
@@ -85,6 +83,11 @@ func (r *RunConfig) flags() []cli.Flag {
 			Usage:       "Do not delete the temp directory after the run",
 			Destination: &r.RetainTempDir,
 		},
+		&cli.StringFlag{
+			Name:        "prepared-dir",
+			Usage:       "Relative directory already prepared. Cannot include version with this.",
```
bergundy, line 88: Could add a comment that in the `version` help string that it's mutually exclusive with `prepared-dir` too.

cmd/run.go
```diff
@@ -113,23 +116,30 @@ func NewRunner(config RunConfig) *Runner {
...
 
 	// Cannot generate history if a version isn't provided explicitly
 	if r.config.GenerateHistory && r.config.Version == "" {
 		return fmt.Errorf("must have explicit version to generate history")
 	}
 
+	// If prepared dir given, validate and make absolute
+	if r.config.Dir != "" {
+		if strings.ContainsAny(r.config.Dir, `\/`) {
+			return fmt.Errorf("prepared directory must not have path separators, it is always relative to the SDK features root")
```
bergundy, line 132: No subdirectories?

cmd/run_java.go
```diff
@@ -49,14 +44,40 @@ dependencies {
...
+	// Create temp dir if needed and prepare the project
+	if r.config.Dir == "" {
+		var err error
+		if r.config.Dir, err = os.MkdirTemp(r.rootDir, "sdk-features-java-test-"); err != nil {
+			return fmt.Errorf("failed creating temp dir: %w", err)
+		}
+		r.createdTempDir = &r.config.Dir
+
+		// Prepare the project but don't build since it'll happen when we run later
+		if err := NewPreparer(r.config.PrepareConfig).PrepareJavaExternal(ctx, false); err != nil {
```
bergundy, line 76: Why not build here? Does it slow things down? Just curious..

cmd/run_python.go
```diff
@@ -67,19 +55,42 @@ sdk-features = { path = "../" }
...
+func (r *Runner) RunPythonExternal(ctx context.Context, run *cmd.Run) error {
+	// Create temp directory and prepare if using specific version
+	if r.config.Version != "" && r.config.Dir == "" {
+		// Create a temp directory and prepare
+		var err error
+		if r.config.Dir, err = os.MkdirTemp(r.rootDir, "sdk-features-python-test-"); err != nil {
+			return fmt.Errorf("failed creating temp dir: %w", err)
+		}
+		r.createdTempDir = &r.config.Dir
+		if err := NewPreparer(r.config.PrepareConfig).PrepareTypeScriptExternal(ctx); err != nil {
```
bergundy, line 83: Typo, should be Python

bergundy (approved):
Mostly nits and one typo found

cmd/run.go
cretz: I think we're filling up these CLI arg docs with too much info. Note how I also don't tell the format of version for each language too. But I'll do it. However, keep in mind we need to be keeping these brief.

cmd/run.go
cretz: No. Right now all of the harnesses are expected to be in `../harness`. We can fix that later, but for now the temporary directories and the prepared directories are expected to be just beneath the sdk-features directory.

cmd/run_java.go
cretz: Because we don't use the build. The "build" is just to prime the system caches. There is no reason, due to how quick Java is, to rework this code to use the built JAR instead of just using `gradle run`.

PR #112, opened 2022-08-09
Query timeout feature

Sushisource:
## What was changed
Added query timeout feature. There was another one for timeout during processing, but the end result looks identical, so I don't see value in implementing that too.

### DIFF
--- a/features/query/timeout_due_to_no_active_workers/feature.go
+++ b/features/query/timeout_due_to_no_active_workers/feature.go
@@ -1,0 +1,47 @@
+package timeout_due_to_no_active_workers
+
+import (
+	"context"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+	"time"
+)
+
+var Feature = harness.Feature{
+	Workflows: Workflow,
+	CheckResult: func(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
+		// Shut off the worker
+		r.Worker.Stop()
+
+		oneSecCtx, cf := context.WithDeadline(ctx, time.Now().Add(time.Second))
+		defer cf()
+		_, err := r.Client.QueryWorkflow(oneSecCtx, run.GetID(), run.GetRunID(), "somequery")
+		r.Require.Error(err)
+		_, ok := err.(*serviceerror.DeadlineExceeded)
+		r.Require.True(ok)
+
+		err = r.RestartWorker()
+		if err != nil {
+			return err
+		}
+
+		// Finish the wf
+		err = r.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), "finish", nil)
+		r.Require.NoError(err)
+		return r.CheckResultDefault(ctx, run)
+	},
+}
+
+func Workflow(ctx workflow.Context) error {
+	err := workflow.SetQueryHandler(ctx, "somequery", func() (bool, error) { return true, nil })
+	if err != nil {
+		return err
+	}
+
+	sigChan := workflow.GetSignalChannel(ctx, "finish")
+	sigChan.Receive(ctx, nil)
+
+	return nil
+}
--- a/features/query/timeout_due_to_no_active_workers/feature.go
+++ b/features/query/timeout_due_to_no_active_workers/feature.go
@@ -13,7 +13,7 @@ var Feature = harness.Feature{
 	Workflows: Workflow,
 	CheckResult: func(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
 		// Shut off the worker
-		r.Worker.Stop()
+		r.StopWorker()
 
 		oneSecCtx, cf := context.WithDeadline(ctx, time.Now().Add(time.Second))
 		defer cf()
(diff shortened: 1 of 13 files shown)


features/query/timeout_due_to_no_active_workers/feature.go
```diff
@@ -0,0 +1,47 @@
...
+		_, ok := err.(*serviceerror.DeadlineExceeded)
+		r.Require.True(ok)
+
+		err = r.RestartWorker()
+		if err != nil {
+			return err
+		}
+
+		// Finish the wf
+		err = r.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), "finish", nil)
```
cretz, line 31: Just let the workflow finish immediately and run `CheckResultDefault` before you even start querying. Might as well run the queries on a closed workflow, you don't have to restart the worker.

harness/go/harness/runner.go
```diff
@@ -304,6 +294,24 @@ func (r *Runner) CheckAssertion(result bool) error {
 	return nil
 }
 
+func (r *Runner) RestartWorker() error {
```
cretz, line 297: Can just call this `StartWorker`

features/query/timeout_due_to_no_active_workers/feature.go
```diff
@@ -0,0 +1,47 @@
...
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+	"time"
+)
+
+var Feature = harness.Feature{
+	Workflows: Workflow,
+	CheckResult: func(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
+		// Shut off the worker
+		r.Worker.Stop()
```
cretz, line 16: I think it'd be clearer to have a helper to stop the worker (e.g. `r.StopWorker()`) that would properly mark it stopped by setting the worker to nil and not try to stop at the end if it was already stopped (i.e. nil). Also `RestartWorker` (which I propose changing to `StartWorker`) will fail on a non-nil worker (i.e. already running).

harness/go/harness/runner.go
```diff
@@ -304,6 +294,24 @@ func (r *Runner) CheckAssertion(result bool) error {
...
+	r.Worker = worker.New(r.Client, r.RunnerConfig.TaskQueue, r.Feature.WorkerOptions)
```
cretz, line 298: Mentioned in other comment, but should fail if `r.Worker` is already set and have a `StopWorker` unset it.

harness/python/feature.py
```diff
@@ -75,12 +78,13 @@ async def run(self) -> None:
...
-            workflows=self.feature.workflows,
-            activities=self.feature.activities,
-        )
+        self.worker = self.create_worker()
+        if self.feature.worker_starter:
+            await self.feature.worker_starter(self)
+        else:
+            await self.start_worker_and_check_wf()
+
+    async def start_worker_and_check_wf(self):
```
cretz, line 87:
```suggestion
    async def start_worker_and_check_workflow(self):
```

Would prefer to avoid these Core-like abbreviations (we tend to be inconsistent in our use of them)

harness/python/feature.py
```diff
@@ -57,6 +59,7 @@ class Feature:
     expect_run_result: Optional[Any]
     start: Optional[Callable[[Runner], Awaitable[WorkflowHandle]]]
     check_result: Optional[Callable[[Runner, WorkflowHandle], Awaitable[None]]]
+    worker_starter: Optional[Callable[[Runner], Awaitable[None]]]
```
cretz, line 62: This looks like more than just `worker_starter`, it looks like it is also supposed to run the workflow and check results. 

If you just need to be in control of shutting down the worker manually, maybe we should just change this to:

```python
self.start_worker()
try:
  # Rest of stuff
finally:
  self.stop_worker()
```

And use `worker.run()` and `worker.stop()` and stop using `async with`. And `stop_worker` can unset the worker so it knows not to do it if already done.

features/query/timeout_due_to_no_active_workers/feature.go
Sushisource: True, but, this works and exists now - the restarting functionality is likely to be useful for something else later.

harness/python/feature.py
```diff
@@ -127,3 +124,20 @@ async def check_result(self, handle: WorkflowHandle) -> None:
                     raise TypeError("Unexpected activity error") from err
             else:
                 raise err
+
+    def start_worker(self):
+        """Creates and starts worker with the task queue & workflows/ activities set, if it is not
+        already running"""
+        if self.worker is None:
```
cretz, line 131: IMO this should fail if worker is not `None` but not a big deal

harness/python/feature.py
```diff
@@ -127,3 +124,20 @@ async def check_result(self, handle: WorkflowHandle) -> None:
...
+            self.worker = Worker(
+                self.client,
+                task_queue=self.task_queue,
+                workflows=self.feature.workflows,
+                activities=self.feature.activities,
+            )
+            self.worker._start()
```
cretz, line 138: If using public API, making this function `async def start_worker` and changing this line to `asyncio.create_task(self.worker.run())` would be the way to only use public API. No biggie currently though.

cretz (approved):

harness/python/feature.py
```diff
@@ -75,13 +75,8 @@ async def run(self) -> None:
...
 
         # Run worker
-        self.worker = Worker(
-            self.client,
-            task_queue=self.task_queue,
-            workflows=self.feature.workflows,
-            activities=self.feature.activities,
-        )
-        async with self.worker:
+        self.start_worker()
```
cretz, line 78: You'll want to add `self.worker: Optional[Worker] = None` as the last statement in `__init__`.

Sushisource:
Part of https://github.com/temporalio/sdk-features/pull/112

PR #124, opened 2022-09-03
Add build-image command for Go SDK

bergundy:
- Use temporalite from temporalio org to get non-CGo version of SQLite driver

Checklist:
- [x] settle on label schema and [lint labels](https://github.com/hadolint/hadolint#linting-labels)
- [x] run against cloud namespace
- [x] run in parallel
- [x] set up workflow manual trigger


cmd/build_image.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return i.dockerBuild(ctx, buildConfig{
+		tags:      tags,
+		labels:    map[string]string{"SDK_REPO_URL": i.config.RepoURL, "SDK_REPO_REF": repoRef},
+		buildArgs: map[string]string{"SDK_VERSION": repoDir, "REPO_DIR_OR_PLACEHOLDER": repoDir},
+	})
+}
+
+func (i *ImageBuilder) buildFromVersion(ctx context.Context) error {
+	version := semver.Canonical(i.config.Version)
```
cretz, line 162: I don't think you should validate this and I don't think you should use the Go library which requires a `v` prefix. Can't we just use the version as the tag directly? We intentionally don't validate version in other SDK features commands.

cmd/build_image.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	for k, v := range config.labels {
+		args = append(args, "--label", fmt.Sprintf("%s=%s", k, v))
+	}
+	for k, v := range config.buildArgs {
+		args = append(args, "--build-arg", fmt.Sprintf("%s=%s", k, v))
+	}
+	args = append(args, i.rootDir)
+
+	i.log.Info("building docker image", tag.NewAnyTag("args", args))
```
cretz, line 229: Should capitalize logs IMO

cretz (approved):
Looks good.

Should be noted that you are still using the internet to download Go dependencies when run. So the need for a version-specific docker build makes less sense for Go than it does for languages where there's actually a pre-build step.

cmd/build_image.go
bergundy: I need this for generating the different semver tags (1, 1.16, 1.16.0), All of our SDKs support semver except for Python for now (granted the `v` prefix is go specific).
I will probably end up changing this a bit when I implement building containers for another language.

bergundy:
> Looks good.
> 
> Should be noted that you are still using the internet to download Go dependencies when run. So the need for a version-specific docker build makes less sense for Go than it does for languages where there's actually a pre-build step.

Hmmm.. what makes you say that? I can run the container with no internet access (turned wifi off on my laptop).
Tests also start immediately and nothing build-related is logged.

cmd/build_image.go
cretz: Ah, Docker tags? :+1:

cretz:
> I can run the container with no internet access (turned wifi off on my laptop).

Hrmm, Unless you are copying over the go mod cache, it should be reaching out to the internet to download dependencies. But I may be confused on what exactly is being downloaded when.

.github/workflows/docker-images.yaml
```diff
@@ -0,0 +1,82 @@
...
+      TEMPORAL_CLOUD_ADDRESS: sdk-ci.a2dd6.tmprl.cloud:7233
+      TEMPORAL_CLOUD_NAMESPACE: sdk-ci.a2dd6
+      REPO_URL: ${{ github.event.pull_request.head.repo.html_url }}
+    steps:
+      - name: Print build information
+        run: 'echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}'
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.17'
```
cretz, line 45: Should bump up to at least 1.18 (1.17 is EOL, I need to fix in other places)

cmd/build_image.go
bergundy: Yup

cretz (approved):

.github/workflows/docker-images.yaml
bergundy: I'll fix in one of the next PRs.

bergundy:
Python tests are still broken, force merged.

PR #132, opened 2022-09-13
Null converter feature test

Quinn-With-Two-Ns:
Add a test for the null data converter.

I renamed the folder because `null` is a keyword in java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Prerequisites:
 - [JDK](https://adoptium.net/?variant=openjdk11&jvmVariant=hotspot) 11+
 - [Node](https://nodejs.org) 16+
 - [Python](https://www.python.org/) 3.10+
-  - [Poetry](https://python-poetry.org/)
+  - [Poetry](https://python-poetry.org/): `poetry install`
   - `setuptools`: `python -m pip install -U setuptools`
 
 Command:
--- a/features/data_converter/empty/feature.go
+++ b/features/data_converter/empty/feature.go
@@ -1,0 +1,81 @@
+package empty
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path"
+	"time"
+
+	"github.com/gogo/protobuf/jsonpb"
+	"go.temporal.io/api/common/v1"
+	historyProto "go.temporal.io/api/history/v1"
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/sdk/workflow"
+)
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	Activities:  Activity,
+	CheckResult: CheckResult,
+}
+
+// run a workflow that calls an activity with a null parameter.
+func Workflow(ctx workflow.Context) error {
+	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 1 * time.Minute,
+	})
+
+	return workflow.ExecuteActivity(ctx, Activity, nil).Get(ctx, nil)
+}
+
+func Activity(ctx context.Context, input *string) error {
+	// check the null input is serialized correctly
+	if input != nil {
+		return temporal.NewNonRetryableApplicationError("Activity input should be nil", "BadResult", nil)
+	}
+	return nil
+}
+
+func CheckResult(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) error {
+	// verify client result is null
+	var result interface{}
+	if err := run.Get(ctx, &result); err != nil {
+		return err
+	}
+	runner.Require.Nil(result)
+
+	// get result payload of ActivityTaskScheduled event from workflow history
+	history := runner.Client.GetWorkflowHistory(ctx, run.GetID(), "", false, 0)
+	event, err := harness.FindEvent(history, func(ev *historyProto.HistoryEvent) bool {
+		attrs := ev.GetActivityTaskScheduledEventAttributes()
+		return attrs != nil
+	})
+	if err != nil {
+		return err
+	}
+
+	attrs := event.GetActivityTaskScheduledEventAttributes()
+	if attrs == nil {
+		return errors.New("could not locate WorkflowExecutionCompleted event")
+	}
+	// verify the activity input payload
+	payload := attrs.GetInput().GetPayloads()[0]
+
+	// load JSON payload from `./payload.json` and compare it to result payload
+	file, err := os.Open(path.Join(runner.Feature.AbsDir, "payload.json"))
+	if err != nil {
+		return err
+	}
+
+	expectedPayload := &common.Payload{}
+	unmarshaler := jsonpb.Unmarshaler{}
+	err = unmarshaler.Unmarshal(file, expectedPayload)
+	if err != nil {
+		return err
+	}
+	runner.Require.Equal(expectedPayload, payload)
+	return nil
+}
--- a/features/data_converter/empty/feature.java
+++ b/features/data_converter/empty/feature.java
@@ -1,0 +1,65 @@
+package data_converter.empty;
+
+import io.temporal.api.common.v1.Payload;
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityMethod;
+import io.temporal.failure.ApplicationFailure;
+import io.temporal.sdkfeatures.Feature;
+import io.temporal.sdkfeatures.Run;
+import io.temporal.sdkfeatures.Runner;
+import io.temporal.sdkfeatures.SimpleWorkflow;
+
+import java.time.Duration;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import com.google.protobuf.util.JsonFormat;
+
+@ActivityInterface
+public interface feature extends Feature, SimpleWorkflow {
+
+  @ActivityMethod
+  void activity(String input);
+
+  class Impl implements feature {
+    /**
+     * run a workflow that calls an activity with a null parameter.
+     */
+    @Override
+    public void workflow() {
+      var activities = activities(feature.class, builder -> builder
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -25,7 +25,7 @@ Prerequisites:
 - [JDK](https://adoptium.net/?variant=openjdk11&jvmVariant=hotspot) 11+
 - [Node](https://nodejs.org) 16+
 - [Python](https://www.python.org/) 3.10+
-  - [Poetry](https://python-poetry.org/)
+  - [Poetry](https://python-poetry.org/): `poetry install`
```
bergundy, line 28: `poetry install` will install this project's dependencies (if run in the proper folder), I'd clarify this in the README

features/data_converter/empty/feature.go
```diff
@@ -0,0 +1,55 @@
...
+func CheckResult(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) error {
+	// verify client result is null
+	var result interface{}
+	if err := run.Get(ctx, &result); err != nil {
+		return err
+	}
+	if result != nil {
+		return fmt.Errorf("invalid result: %v", result)
+	}
+	history := runner.Client.GetWorkflowHistory(ctx, run.GetID(), "", false, 0)
```
bergundy, line 33: We could probably abstract this even further and add a method to get the completion event for a workflow (applies to all lang harnesses)

bergundy (commented):
I would still like to test the null payload converter in Java and Go, we didn't end up testing those.

features/data_converter/empty/feature.go
cretz: I still am not convinced we need to check payloads. IMO, we only need to check end user behavior. If it comes back as nil, that's fine. (but I understand others disagree w/ me here)

features/data_converter/empty/feature.go
```diff
@@ -0,0 +1,55 @@
...
+		if attrs != nil {
+			break
+		}
+	}
+	if attrs == nil {
+		return errors.New("could not locate WorkflowExecutionCompleted event")
+	}
+
+	// verify the workflow execution completed event has no payload
+	runner.Require.Empty(attrs.GetResult().GetPayloads())
```
cretz, line 53: If you use `runner.Require.Empty` here, you might as well use `runner.Require.NotNil` for your result check above.

features/data_converter/empty/feature.go
cretz: I still am not convinced we need to check payloads. IMO, we only need to check end user behavior. If it comes back as nil, that's fine. (but I understand others disagree w/ me here)

features/data_converter/empty/feature.py
```diff
@@ -0,0 +1,50 @@
...
+    e = next(
+        e
+        for e in history
+        if e.event_type == EventType.EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED
+    )
+    payload = e.workflow_execution_completed_event_attributes.result.payloads[0]
+
+    # load JSON payload from `./payload.json` and compare it to JSON representation of result payload
+    with open(
+        os.path.join(os.path.dirname(runner.feature.file), "payload.json"),
```
cretz, line 39: Is it really worth an entire file just to check empty data and a specific metadata value?

cretz (approved):
Awesome!

Note, this repo is young, so if you're finding helpers that can be used across tests feel free to add them to harnesses/utils.

features/data_converter/empty/feature.py
```diff
@@ -0,0 +1,65 @@
...
+from temporalio.api.common.v1 import Payload
+from temporalio.api.enums.v1 import EventType
+from temporalio.client import WorkflowHandle
+from temporalio.exceptions import ApplicationError
+
+from harness.python.feature import Runner, register_feature
+
+
+@activity.defn
+async def null_activity(input: str) -> None:
```
bergundy, line 15:
```suggestion
async def null_activity(input: Optional[str]) -> None:
```

features/data_converter/empty/feature.go
```diff
@@ -0,0 +1,74 @@
...
+	// verify client result is null
+	var result interface{}
+	if err := run.Get(ctx, &result); err != nil {
+		return err
+	}
+	runner.Require.Nil(result)
+
+	// get result payload of ActivityTaskScheduled event from workflow history
+	history := runner.Client.GetWorkflowHistory(ctx, run.GetID(), "", false, 0)
+	event, err := harness.FindEvent(history, func(ev *historyProto.HistoryEvent) bool {
```
bergundy, line 51: Check if err is not nil

features/data_converter/empty/feature.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// run a workflow that calls an activity with a null parameter.
+func Workflow(ctx workflow.Context) error {
+	ctx = workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ScheduleToCloseTimeout: 1 * time.Minute,
+	})
+
+	err := workflow.ExecuteActivity(ctx, Activity, nil).Get(ctx, nil)
+	return err
```
bergundy, line 30: nit: can condense this into a single line

features/data_converter/empty/feature.java
```diff
@@ -0,0 +1,65 @@
...
+  void activity(String input);
+
+  class Impl implements feature {
+    /**
+     * run a workflow that calls an activity with a null parameter.
+     */
+    @Override
+    public void workflow() {
+      var activities = activities(feature.class, builder -> builder
+              .setScheduleToCloseTimeout(Duration.ofMinutes(1)));
```
bergundy, line 34: We typically use `startToClose`, while schedule to close is fine in this specific case, it's not recommended in most cases where activities should retry indefinitely.
Once a schedule to close timeout is hit the activity will no longer be retried.

The reason I mention this is because this repo has some value as "samples" or "reference" repo where users might look up how the system behaves, we want people to copy our recommended best practices.

features/data_converter/empty/feature.py
bergundy: I think so

bergundy: If we had cross SDK testing capabilities, I'd say we don't have to test the generated payloads but it's nice to know that null represented the same in all SDKs.

features/data_converter/empty/feature.py
```diff
@@ -0,0 +1,66 @@
...
+from temporalio.exceptions import ApplicationError
+
+from harness.python.feature import Runner, register_feature
+
+
+@activity.defn
+async def null_activity(input: Optional[str]) -> None:
+    # check the null input is serialized correctly
+    if input != None:
+        raise ApplicationError("Activity input should be None")
```
bergundy, line 19: Is this non-retryable?

bergundy: nit: All of the application failures in this feature _should_ be non-retryable to prevent needless retries.

features/data_converter/empty/feature.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Feature } from '@temporalio/harness';
+import * as assert from 'assert';
+import { ApplicationFailure } from '@temporalio/common';
+import expectedPayload from './payload.json';
+import * as wf from '@temporalio/workflow';
+
+const activitiesImpl = {
+  async activity(input: any): Promise<void> {
+    if (input != undefined) {
+      throw new ApplicationFailure('Activity input should be undefined', null, true, undefined, undefined);
```
bergundy, line 11: You can just throw an `Error` for retryable or use the `ApplicationFailure.nonRetryable` factory method instead of the constructor.

harness/go/harness/util.go
```diff
@@ -13,3 +15,17 @@ var RetryDisabled = &temporal.RetryPolicy{MaximumAttempts: 1}
 func AppErrorf(msg string, args ...interface{}) error {
 	return temporal.NewApplicationError(fmt.Sprintf(msg, args...), "SdkFeaturesError")
 }
+
+// Find the first event in the history that meets the condition.
+func FindEvent(history client.HistoryEventIterator, cond func(*historypb.HistoryEvent) bool) (*historypb.HistoryEvent, error) {
```
bergundy, line 20: Nice, now that we have this, can you rewrite the binary test to use it?

bergundy (approved):
Nice, I didn't have too many comments.
Mostly nits, we can merge after you go over this round of comments AFAIC.

harness/go/harness/util.go
Quinn-With-Two-Ns: yep ill add a commit it do that.

features/data_converter/empty/feature.py
Quinn-With-Two-Ns: no reason for it to be retryable, I'll update

features/data_converter/empty/feature.java
Quinn-With-Two-Ns: will do

features/data_converter/empty/feature.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Feature } from '@temporalio/harness';
+import * as assert from 'assert';
+import { ApplicationFailure } from '@temporalio/common';
+import expectedPayload from './payload.json';
+import * as wf from '@temporalio/workflow';
+
+const activitiesImpl = {
+  async activity(input: any): Promise<void> {
+    if (input != undefined) {
+      throw ApplicationFailure.nonRetryable('Activity input should be undefined', 'BadResult');
```
bergundy, line 11: nit: the type is optional and `BadResult` is not really a good name for the error.
Not blocking this PR.

features/data_converter/empty/feature.ts
```diff
@@ -0,0 +1,44 @@
+import { JSONToPayload } from '@temporalio/common/lib/proto-utils';
+import { Feature } from '@temporalio/harness';
+import * as assert from 'assert';
+import { ApplicationFailure } from '@temporalio/common';
+import expectedPayload from './payload.json';
+import * as wf from '@temporalio/workflow';
+
+const activitiesImpl = {
+  async activity(input: any): Promise<void> {
```
bergundy, line 9: nit: This could be an optional parameter or union with undefined or just `undefined`.

```suggestion
  async activity(input?: string): Promise<void> {
```

bergundy (commented):
This is good to merge, I had a couple more nits but they can be ignored.

PR #135, opened 2022-09-21
Continue as new feature test

Quinn-With-Two-Ns:
Test compliance between the different SDKs when using continue as new

### DIFF
--- a/features/continue_as_new/continue_as_same/feature.go
+++ b/features/continue_as_new/continue_as_same/feature.go
@@ -1,0 +1,68 @@
+package continue_as_same
+
+import (
+	"context"
+	"time"
+
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	InputData  = "InputData"
+	MemoKey    = "MemoKey"
+	MemoValue  = "MemoValue"
+	WorkflowID = "TestID"
+)
+
+var Feature = harness.Feature{
+	Workflows: Workflow,
+	CheckResult: func(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
+		var result string
+		if err := run.Get(ctx, &result); err != nil {
+			return err
+		}
+		r.Require.Equal(InputData, result)
+
+		execution, err := r.Client.DescribeWorkflowExecution(ctx, run.GetID(), run.GetRunID())
+		if err != nil {
+			return err
+		}
+		// Workflow ID does not change after continue as new
+		r.Require.Equal(WorkflowID, run.GetID())
+		// Memos do not change after continue as new
+		memo := execution.GetWorkflowExecutionInfo().GetMemo().GetFields()
+		r.Require.NotNil(memo)
+		memoPayload, ok := memo[MemoKey]
+		r.Require.True(ok)
+		var testMemo string
+		err = converter.GetDefaultDataConverter().FromPayload(memoPayload, &testMemo)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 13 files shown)


features/continue_as_new/continue_as_same/feature.go
```diff
@@ -0,0 +1,72 @@
...
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	InputData  = "InputData"
+	MemoKey    = "MemoKey"
+	MemoValue  = "MemoValue"
+	WorkflowID = "TestID"
+)
```
cretz, line 18: Since these are test cases, you can just inline the literals if you want

features/continue_as_new/continue_as_same/feature.go
```diff
@@ -0,0 +1,72 @@
...
+			Memo: map[string]interface{}{
+				MemoKey: MemoValue,
+			},
+			WorkflowExecutionTimeout: 1 * time.Minute,
+		}
+		run, err := runner.Client.ExecuteWorkflow(ctx, opts, runner.Feature.Workflows[0], InputData)
+		if err != nil {
+			return nil, err
+		}
+		return run, nil
```
cretz, line 61:
```suggestion
		return runner.Client.ExecuteWorkflow(ctx, opts, runner.Feature.Workflows[0], InputData)
```

features/continue_as_new/continue_as_same/feature.py
```diff
@@ -0,0 +1,52 @@
...
+            return input
+        workflow.continue_as_new(arg=input)
+
+
+async def start(runner: Runner) -> WorkflowHandle:
+    defn = workflow._Definition.must_from_class(runner.feature.workflows[0])
+    return await runner.client.start_workflow(
+        defn.name,
+        id=WORKFLOW_ID,
+        arg=INPUT_DATA,
```
cretz, line 28:
```suggestion
    return await runner.client.start_workflow(
        Workflow,
        INPUT_DATA,
        id=WORKFLOW_ID,
```

cretz (approved):

features/continue_as_new/continue_as_same/feature.go
Quinn-With-Two-Ns: 🤦

Quinn-With-Two-Ns:
looks like upgrading python sdk version broke some other stuff I gotta fix

features/continue_as_new/continue_as_same/feature.go
```diff
@@ -0,0 +1,72 @@
...
+		r.Require.NotNil(memo)
+		memoPayload, ok := memo[MemoKey]
+		r.Require.True(ok)
+		var testMemo string
+		err = converter.GetDefaultDataConverter().FromPayload(memoPayload, &testMemo)
+		if err != nil {
+			return err
+		}
+		r.Require.Equal(MemoValue, testMemo)
+		return nil
```
bergundy, line 46: All of this isn't necessary IMO, it's enough to check the memo from the workflow

features/continue_as_new/signal/README.md
```diff
@@ -1,4 +1,4 @@
-# Continue as new
+# signal
```
bergundy, line 1: Maybe call this signals_block_completion.
I don't know if I'd put this test in continue-as-new, it's any workflow completion that's blocked by unprocessed signals coming in while the concluding task is processed by a worker.

bergundy (approved):
The test could be simplified to check the memo in workflow code but otherwise mostly LGTM.

cretz:
@Quinn-With-Two-Ns - Yeah, sorry, there were slight backwards incompatible changes, see https://github.com/temporalio/sdk-python/releases/tag/0.1b2

PR #147, opened 2022-10-20
Finish Query Features

Sushisource:
Finishes up the remaining query features. There's a couple todos about things that we might address in the sdks themselves

### DIFF
--- a/features/query/unexpected_arguments/feature.go
+++ b/features/query/unexpected_arguments/feature.go
@@ -1,0 +1,53 @@
+package unexpected_arguments
+
+import (
+	"context"
+	"fmt"
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+var Feature = harness.Feature{
+	Workflows: Workflow,
+	CheckResult: func(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
+		// Go does not reject anything
+
+		var qRes string
+
+		q, err := r.Client.QueryWorkflow(ctx, run.GetID(), run.GetRunID(), "theQuery", 123)
+		r.Require.NoError(err)
+		r.Require.NoError(q.Get(&qRes))
+		r.Require.Equal("got 123", qRes)
+
+		// Drops extra arg
+		q, err = r.Client.QueryWorkflow(ctx, run.GetID(), run.GetRunID(), "theQuery", 123, true)
+		r.Require.NoError(err)
+		r.Require.NoError(q.Get(&qRes))
+		r.Require.Equal("got 123", qRes)
+
+		// Assumes default value
+		q, err = r.Client.QueryWorkflow(ctx, run.GetID(), run.GetRunID(), "theQuery")
+		r.Require.NoError(err)
+		r.Require.NoError(q.Get(&qRes))
+		r.Require.Equal("got 0", qRes)
+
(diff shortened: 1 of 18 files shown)


features/query/unexpected_arguments/feature.py
```diff
@@ -0,0 +1,52 @@
...
+    def the_query(self, arg: int):
+        return f"hi {arg}"
+
+    @workflow.signal(name="finish")
+    def finish_sig(self):
+        self.be_done = True
+
+
+async def checker(_: Runner, handle: WorkflowHandle):
+    # Wrong type TODO: Should be rejected
```
cretz, line 29: Try a string

features/query/unexpected_query_type_name/feature.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @workflow.signal(name="finish")
+    def finish_sig(self):
+        self.be_done = True
+
+
+async def checker(_: Runner, handle: WorkflowHandle):
+    try:
+        await handle.query("nonexistent")
+    except RPCError:
```
cretz, line 27: Opened https://github.com/temporalio/sdk-python/issues/162

features/query/unexpected_query_type_name/feature.py
```diff
@@ -0,0 +1,36 @@
...
+        pass
+    else:
+        raise Exception("Query with nonexistent handler must fail")
```
cretz, line 30:
```suggestion
    try:
        await handle.query("nonexistent")
        raise Exception("Query with nonexistent handler must fail")
    except RPCError:
        pass
```

If you want to save a line :-)

features/query/unexpected_arguments/feature.go
```diff
@@ -0,0 +1,53 @@
...
+func Workflow(ctx workflow.Context) error {
+	err := workflow.SetQueryHandler(ctx, "theQuery", func(a int) (string, error) {
+		return fmt.Sprintf("got %d", a), nil
+	})
+	if err != nil {
+		return err
+	}
+
+	sigChan := workflow.GetSignalChannel(ctx, "finish")
+	sigChan.Receive(ctx, nil)
```
cretz, line 50: If you wanted to you could just let the workflow finish and let all these queries you're asserting run on completed workflows. Sure it costs a bit in cache/replay, but meh. Of course signal-based finish is fine too.

features/query/unexpected_return_type/feature.py
```diff
@@ -0,0 +1,47 @@
...
+        return None
+
+    @workflow.query(name="q")
+    def num_q(self) -> int:
+        return 1
+
+
+async def checker(_: Runner, handle: WorkflowHandle):
+    q1 = await handle.query(FakeWf.num_q)
+    # Python (at the moment) doesn't detect any type mismatch here
```
cretz: Opened https://github.com/temporalio/sdk-python/issues/163

cretz (approved):

features/query/unexpected_return_type/feature.py
cretz: Nevermind

features/query/unexpected_arguments/feature.py
Sushisource: Yeah, it'll work with a string. Keeping this one in so we can see it get fixed tho :)

cretz: I'm not sure it will get fixed. If Python typing considers `True` an acceptable value for `int`, so do I.

PR #152, opened 2022-11-10
Build Typescript Docker Images

Sushisource:
Fix docker image building, and pass TLS settings through to the TS harness

### DIFF
--- a/cmd/build_image.go
+++ b/cmd/build_image.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -144,8 +145,7 @@ func (i *ImageBuilder) buildFromRepo(ctx context.Context) error {
 		return err
 	}
 
-	// Not using filepath.Join because we need POSIX path on windows
-	repoDir := fmt.Sprintf("%s/%s", filepath.Base(tempDir), repoBaseDir)
+	repoDir := path.Join(filepath.Base(tempDir), repoBaseDir)
 
 	// Get the actual ref (in case passed in ref is a branch name or tag)
 	repoRef, err := i.gitRef(ctx, filepath.Join(tempDir, repoBaseDir, ".git"))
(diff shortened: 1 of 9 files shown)


cmd/build_image.go
```diff
@@ -143,7 +144,8 @@ func (i *ImageBuilder) buildFromRepo(ctx context.Context) error {
 		return err
 	}
 
-	repoDir := filepath.Join(filepath.Base(tempDir), repoBaseDir)
+	// Not using filepath.Join because we need POSIX path on windows
```
cretz, line 147: This is what `path.Join` is for

bergundy: Oh nice, I didn't know that.

cmd/build_image.go
```diff
@@ -159,7 +161,7 @@ func (i *ImageBuilder) buildFromRepo(ctx context.Context) error {
 			"io.temporal.sdk.repo-url": i.config.RepoURL,
 			"io.temporal.sdk.repo-ref": repoRef,
 		},
-		buildArgs: map[string]string{"SDK_VERSION": repoDir, "REPO_DIR_OR_PLACEHOLDER": repoDir},
+		buildArgs: map[string]string{"SDK_VERSION": fmt.Sprintf("/app/%s", repoDir), "REPO_DIR_OR_PLACEHOLDER": repoDir},
```
bergundy, line 164: Might as well use `path.Join` here too then.

cmd/run_typescript.go
```diff
@@ -127,10 +143,35 @@ func (r *Runner) RunTypeScriptExternal(ctx context.Context, run *cmd.Run) error
...
+	runArgs := []string{
+		"-r",
+		"tsconfig-paths/register",
+		"./tslib/harness/ts/main.js",
+		"--server",
+		r.config.Server,
+		"--namespace",
+		r.config.Namespace,
+		"--client-key-path",
+		r.config.ClientKeyPath,
```
bergundy, line 155: Looks lie we provide this below

bergundy (approved):

cmd/run_typescript.go
Sushisource: Thanks, forgot to remove that.

PR #154, opened 2022-11-14
Update Go SDK and remove server/temporalite dependency

cretz:
## What was changed

* Change Temporalite to be downloaded and run as a sub process instead of embedded
* Change all logs to use Go SDK logging abstraction instead of Go server logging abstraction
* Remove server and Temporalite dependencies (this frees this repo from waiting on server releases)
* Update Go SDK to v1.18.1


.github/workflows/ci.yaml
```diff
@@ -107,6 +107,6 @@ jobs:
       lang: go
       # Latest release at the time this was written
       # TODO(bergundy): Find some way to automatically upgrade to "latest"
-      sdk-version: 'v1.16'
+      sdk-version: 'v1.18.1'
```
cretz, line 110: This was required because v1.16.0 no longer works with the latest API repo due to:

> go.temporal.io/api/update/v1: module go.temporal.io/api@latest found (v1.13.0), but does not contain package go.temporal.io/api/update/v1

v1.16.0 and v1.17.0 of the Go SDK are not compatible with latest API module because we had update code in there but have since deleted that API package.

harness/go/temporalite/temporalite.go
```diff
@@ -0,0 +1,245 @@
...
+	// Temporalite has no way to give us the port they chose, so we have to find
+	// our own free port
+	port, err := options.GetFrontendPort()
+	if err != nil {
+		return nil, fmt.Errorf("failed getting free port: %w", err)
+	}
+	portStr := strconv.Itoa(port)
+
+	// Start
+	args := []string{"start", "--ephemeral", "--headless", "--namespace", options.Namespace, "--port", portStr}
```
bergundy, line 88: Might as well add `--dynamic-config-value system.forceSearchAttributesCacheRefreshOnRead=true --dynamic-config-value system.enableActivityEagerExecution=true`

cmd/run.go
```diff
@@ -178,25 +178,18 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
...
-			// TODO(cretz): Allow multiple?
-			temporalite.WithNamespaces(r.config.Namespace),
-			temporalite.WithPersistenceDisabled(),
-			temporalite.WithDynamicPorts(),
-			// TODO(cretz): Server is noisy, allow separate config
-			// temporalite.WithLogger(r.log),
-			temporalite.WithLogger(log.NewNoopLogger()),
-		)
+		server, err := temporalite.Start(temporalite.Options{
+			// Log: r.log,
```
bergundy, line 182: leftover comment?

bergundy (approved):

harness/go/temporalite/temporalite.go
cretz: Can do

cretz:
CI is currently breaking because it builds Python with the sandbox and the features are doing side-effecting things in the sandbox which they shouldn't

harness/go/temporalite/temporalite.go
cretz: Done

cmd/run.go
cretz: Done

cretz:
I am merging anyways. Python tests here have been broken ever since I made breaking changes since the Python tests against `main`. Hopefully a new release will alleviate this.

PR #156, opened 2022-11-14
Java docker image

Sushisource:


.github/workflows/ci.yaml
```diff
@@ -137,14 +137,16 @@ jobs:
...
-  build-java-master-docker-images:
-    uses: ./.github/workflows/docker-images.yaml
-    secrets: inherit
-    with:
-      lang: java
-      sdk-repo-ref: master
-      sdk-repo-url: https://github.com/temporalio/sdk-java
-      run-tag: master
+# TODO: Need PR to add some gradle stuff in sdk-java to simplify this, but also we're not going
+#     to be running this as a regular part of CI anyway.
```
cretz, line 141: :+1: For not making this a regular part of CI. We should move these out of `ci.yaml` into a `workflow_dispatch` GH workflow. See https://github.com/temporalio/sdk-python/blob/main/.github/workflows/run-bench.yml for an example. this means you can _manually_ start a workflow by going to its action, e.g. https://github.com/temporalio/sdk-python/actions/workflows/run-bench.yml. You get a nice form and everything. You could make all of these options dropdowns or other fields.

(I personally disagree w/ per SDK docker images anyways for reasonably fast building SDKs; there's no reason an sdk features container can't download what it needs live)

Sushisource: Yes, that will all be done in another PR.

Having the images as baked as possible is a good thing though, especially when we're spinning up many workers repeatedly for load tests, etc.

.github/workflows/ci.yaml
```diff
@@ -1,10 +1,15 @@
...
+
 on: # rebuild any PRs and main branch changes
   pull_request:
   push:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
bergundy, line 11: What's this fanciness? Didn't know they had this.

.github/workflows/ci.yaml
bergundy: We're going to be using these containers in many places, baking different SDK versions.
Having to redo all of the work the build steps do here when we use the containers is not a viable option.

bergundy: Based on the recent design, we'll want to trigger master build from Java SDK PRs so it's fine that this is disabled.

.github/workflows/ci.yaml
Sushisource: Yeah pretty important to add to the higher-cost workflows (though, I say that and we're using free runners, but, whatever)

bergundy (approved):

PR #158, opened 2022-11-15
Initial schedule features

cretz:
## What was changed

* Added `schedule/cron` feature in all languages w/ basic cron behavior
* Added `schedule/basic_workflow` for just Go (no other released SDK has scheduling yet)

### DIFF
--- a/features/schedule/basic_workflow/feature.go
+++ b/features/schedule/basic_workflow/feature.go
@@ -1,0 +1,122 @@
+package basic_workflow
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+var Feature = harness.Feature{
+	Workflows: BasicScheduleWorkflow,
+	Execute:   Execute,
+}
+
+func BasicScheduleWorkflow(ctx workflow.Context, arg string) (string, error) {
+	return arg, nil
+}
+
+func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Start schedule every 2s
+	workflowID := uuid.NewString()
+	handle, err := r.Client.ScheduleClient().Create(ctx, client.ScheduleOptions{
+		// TODO(cretz): Required? https://github.com/temporalio/sdk-go/issues/957
+		ID:      uuid.NewString(),
+		Spec:    client.ScheduleSpec{Intervals: []client.ScheduleIntervalSpec{{Every: 2 * time.Second}}},
+		Overlap: enums.SCHEDULE_OVERLAP_POLICY_BUFFER_ONE,
+		// TODO(cretz): Would prefer if this were a pointer. See
+		// https://github.com/temporalio/sdk-go/issues/958.
+		Action: client.ScheduleWorkflowAction{
+			ID:        workflowID,
+			Workflow:  BasicScheduleWorkflow,
+			Args:      []interface{}{"arg1"},
+			TaskQueue: r.TaskQueue,
+		},
+	})
+	r.Require.NoError(err)
+
+	// Remove the schedule on complete
+	defer func() {
+		if err := handle.Delete(context.Background()); err != nil {
+			r.Log.Warn("Failed deleting schedule handle", "error", err)
+		}
+	}()
+
+	// Confirm simple describe
+	desc, err := handle.Describe(ctx)
+	r.Require.NoError(err)
+	r.Require.Equal(workflowID, desc.Schedule.Action.(client.ScheduleWorkflowAction).ID)
+
+	// Confirm simple list
+	iter, err := r.Client.ScheduleClient().List(ctx, client.ScheduleListOptions{})
+	r.Require.NoError(err)
+	foundSchedule := false
+	for iter.HasNext() && !foundSchedule {
+		e, err := iter.Next()
+		r.Require.NoError(err)
+		foundSchedule = e.ID == handle.GetID()
+	}
+	r.Require.True(foundSchedule)
+
+	// Wait for first completion
+	waitCompletedWith(ctx, r, workflowID, "arg1")
+
+	// Update and change arg
+	err = handle.Update(ctx, client.ScheduleUpdateOptions{
+		DoUpdate: func(in client.ScheduleUpdateInput) (*client.ScheduleUpdate, error) {
+			update := &client.ScheduleUpdate{Schedule: &in.Description.Schedule}
+			action := update.Schedule.Action.(client.ScheduleWorkflowAction)
+			action.Args = []interface{}{"arg2"}
+			update.Schedule.Action = action
+			return update, nil
+		},
+	})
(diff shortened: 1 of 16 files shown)


features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+
+func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Start schedule every 2s
+	workflowID := uuid.NewString()
+	handle, err := r.Client.ScheduleClient().Create(ctx, client.ScheduleOptions{
+		// TODO(cretz): Required? https://github.com/temporalio/sdk-go/issues/957
+		ID:      uuid.NewString(),
+		Spec:    client.ScheduleSpec{Intervals: []client.ScheduleIntervalSpec{{Every: 2 * time.Second}}},
+		Overlap: enums.SCHEDULE_OVERLAP_POLICY_BUFFER_ONE,
+		// TODO(cretz): Would prefer if this were a pointer. See
```
mjameswh, line 33: Issue has been fixed, this need to be a pointer.

features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil, nil
+}
+
+func waitCompletedWith(ctx context.Context, r *harness.Runner, id string, untilResult string) {
+	// Wait a max of 10s, waiting 1s between non-error tries
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	for {
+		// Get all completed
+		req := &workflowservice.ListWorkflowExecutionsRequest{
```
mjameswh, line 93: I would use `DescribeScedule.info.recentActions` instead of listing all workflow executions. Would be much more easier and that would test a little bit more the schedule implementation itself.

features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+	return arg, nil
+}
+
+func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Start schedule every 2s
+	workflowID := uuid.NewString()
+	handle, err := r.Client.ScheduleClient().Create(ctx, client.ScheduleOptions{
+		// TODO(cretz): Required? https://github.com/temporalio/sdk-go/issues/957
+		ID:      uuid.NewString(),
+		Spec:    client.ScheduleSpec{Intervals: []client.ScheduleIntervalSpec{{Every: 2 * time.Second}}},
```
mjameswh, line 31: I see more risks of bugs across SDKs with the structured calendar format than the interval spec format... Can we consider defining this test using an equivalent structured calandar spec format instead? For example (using TS syntax): `second: { start: 0, end: 59, step: 2 }, minute: '*', hour: '*', ...`

features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func BasicScheduleWorkflow(ctx workflow.Context, arg string) (string, error) {
+	return arg, nil
+}
+
+func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Start schedule every 2s
+	workflowID := uuid.NewString()
+	handle, err := r.Client.ScheduleClient().Create(ctx, client.ScheduleOptions{
```
mjameswh, line 28: Starting a workflow every 2 seconds can be pretty aggressive in a "non production level" test environment. Just in case something goes wrong and the schedule stays active longer than expected, I would set `limitedActions = true`, and `remainingActions = 2`, or `endAt = now + 2 minutes`...

features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+		// TODO(cretz): Required? https://github.com/temporalio/sdk-go/issues/957
+		ID:      uuid.NewString(),
```
mjameswh, line 30: nit: I'd suggest following the test harness' naming convention for schedule id (`${this.source.relDir}-${randomUUID()}`), and `${scheduleId}-workflow` for the Workflow ID.

mjameswh (approved):
A few suggestions but overall LGTM

features/schedule/basic_workflow/feature.go
```diff
@@ -0,0 +1,122 @@
...
+
+func BasicScheduleWorkflow(ctx workflow.Context, arg string) (string, error) {
+	return arg, nil
+}
+
+func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Start schedule every 2s
+	workflowID := uuid.NewString()
+	handle, err := r.Client.ScheduleClient().Create(ctx, client.ScheduleOptions{
+		// TODO(cretz): Required? https://github.com/temporalio/sdk-go/issues/957
```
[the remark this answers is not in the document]
cretz: It is not resolved in a released version though. Cannot update.

features/schedule/basic_workflow/feature.go
cretz: It is not resolved in a released version though. Cannot update.

features/schedule/basic_workflow/feature.go
cretz: > following the test harness' naming convention for schedule id 

The harness doesn't have a naming convention for schedule ID. I don't think it matters much the shape of the identifier in these cases so long as unique. Notice the workflow type is the function name not prefixed with its qualified path to help the UI either.

features/schedule/basic_workflow/feature.go
cretz: That sounds like a new test. I can think of dozens and dozens of tests if we need to test each input form. These are the initial basic tests, not meant to validate each field.

features/schedule/basic_workflow/feature.go
cretz: I can add that. I still want to list workflows because I want to test the non-schedule result of the schedule on the actual system too.

features/schedule/basic_workflow/feature.go
cretz: The schedule is only supposed to be valid for the life of this test in an isolated way. I delete the schedule before this returns.

Quinn-With-Two-Ns:
LGTM, assuming there will be another PR once typescript is released?

Quinn-With-Two-Ns (approved):

cretz:
> LGTM, assuming there will be another PR once typescript is released?

Yes

features/schedule/basic_workflow/feature.go
mjameswh: > I can think of dozens and dozens of tests if we need to test each input form. These are the initial basic tests, not meant to validate each field.

I might have not expressed myself clearly here... I totally agree that testing tons of inputs on schedules in features tests would just be insane and would not bring in much value.

My argument was rather that, from a cross SDK point of view, if we have to retain only a very minimal number of test inputs for schedule testing, then it seems to me that there would be more value in testing some structured calandar input (because it is more likely to expose divergence between SDKs) rather than the rather trivial interval format. The specific calandar I proposed is functionally equivalent to the "every 2 seconds" interval, which means that testing methodology remains the same.

But yes, that can definitely be a distinct test... Your call.

cretz: Hrmm, I was going for testing how a user might write "every X interval" schedule.

PR #162, opened 2022-11-16
GHA Workflow Refactoring

Sushisource:
## What was changed
* Add java testing workflow
* Don't target main/master of SDKs from PRs to this repo
* Preparation necessary for triggering from server/sdk repos

## Why?
Great testathon 2022

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,18 +94,13 @@ jobs:
     uses: ./.github/workflows/python.yaml
     with:
       python-repo-path: temporalio/sdk-python
-      # TODO(cretz): "main" is unstable
       python-repo-ref: 0.1b3
 
-  # TODO(cretz): Go master is unstable
-  # build-go-master-docker-images:
-  #   uses: ./.github/workflows/docker-images.yaml
-  #   secrets: inherit
-  #   with:
-  #     lang: go
-  #     sdk-repo-ref: master
-  #     sdk-repo-url: https://github.com/temporalio/sdk-go
-  #     run-tag: master
+  feature-tests-java:
+    uses: ./.github/workflows/java.yaml
+    with:
+      java-repo-path: temporalio/sdk-java
+      java-repo-ref: v1.17.0
 
   build-go-latest-docker-images:
     uses: ./.github/workflows/docker-images.yaml
@@ -118,15 +113,6 @@ jobs:
       semver-latest: major
       run-tag: '1.18.1'
 
-  build-node-main-docker-images:
-    uses: ./.github/workflows/docker-images.yaml
-    secrets: inherit
-    with:
-      lang: ts
-      sdk-repo-ref: main
-      sdk-repo-url: https://github.com/temporalio/sdk-typescript
-      run-tag: main
-
   build-node-latest-docker-images:
     uses: ./.github/workflows/docker-images.yaml
     secrets: inherit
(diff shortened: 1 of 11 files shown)


Sushisource:
All SDKs are able to successfully run these workflows now. https://github.com/temporalio/sdk-go/pull/961 fails b/c there is a compat break, we still have to decide our exact strategy there.

.github/workflows/ci.yaml
```diff
@@ -81,31 +81,26 @@ jobs:
   feature-tests-ts:
     uses: ./.github/workflows/typescript.yaml
     with:
-      typescript-repo-path: temporalio/sdk-typescript
-      typescript-repo-ref: main
+      version: 1.4.4
```
cretz, line 84: I have opened https://github.com/temporalio/sdk-features/issues/166 for `--version latest` support. I may be able to get to it today.

I wonder, even if we add that, would it be ok to put in here? Or are there concerns of the mismatch in time between when an SDK is released and when SDK features is updated to support that release? If we're not gonna put it in here, I might not even work on the issue as it may not be worth it.

cretz (approved):

.github/workflows/ci.yaml
Sushisource: Yeah, that's a good question. I kinda like the explicitness of this.

We could maybe do something like check if the explicit version is _not_ latest and let us know somehow? I'm just not sure what the delivery mechanism for that would be.

cretz: Pros of explicitness:
* No downtime for SDK features because it has to be explicitly upgraded instead of being different

Cons of explicitness:
* If a release doesn't break any SDK features, people will have no reason to update this and it may get stale

Yeah, there's no good notification mechanism that can't be ignored, heh.

Maybe at the least we should pull the version out of what is set in the SDK features repo root so we don't have to remember to update there and here? Something like `--version current`? Also then you could put dependabot on this repo to tell you when to upgrade (surely you can configure it to just be for our SDKs and not all of its other noise).

Would be curious @bergundy's opinion here.

bergundy: +1 for #166, I'd just make sure that we get the concrete version when we build docker containers so we can properly tag the images.

.github/workflows/ci.yaml
```diff
@@ -138,17 +124,6 @@ jobs:
       semver-latest: major
       run-tag: '1.4.4'
 
-# TODO: Need PR to add some gradle stuff in sdk-java to simplify this, but also we're not going
```
bergundy, line 141: While technically we could add this trigger from Java SDK to get master docker builds pushed to a repository I don't think master images are a part of our matrix.

I would want us to test against cloud on SDK main / master but that doesn't require building a docker image.

bergundy: (and shouldn't be defined as a part of this matrix so we're good)

.github/workflows/java.yaml
```diff
@@ -0,0 +1,64 @@
+name: Java Features Testing
```
bergundy, line 1: 🙌

harness/ts/main.ts
```diff
@@ -76,15 +77,17 @@ async function run() {
       });
       await runner.run();
     } catch (err) {
-      console.error(`Feature ${featureDir} failed with ${err}`, (err as any).stack);
+      const errstr = `Feature ${featureDir} failed with ${err}`;
+      failedFeaturesStr += errstr + '\n';
+      console.error(errstr, (err as any).stack);
```
bergundy, line 82: This should do it.
```suggestion
      console.error(`Feature ${featureDir} failed`, err);
```

bergundy (approved):
Nice!
Now that we test all SDKs, do we still need to build them (e.g. build-java)?

.github/workflows/ci.yaml
bergundy: I'd be fine with explicit version too as long as we add a release step in our guidelines to bump here or better yet if there was a good way to automate it (assuming we have time for it).

Sushisource:
> Nice! Now that we test all SDKs, do we still need to build them (e.g. build-java)?

Indeed I was considering blasting those, but they run linters which aren't necessary in the reusable workflow defs. I could write ifs to exclude them or just run them every time... but the isolation maybe makes things a touch more readable. :man_shrugging: 

We can always get rid of them later if it's a problem.

harness/ts/main.ts
Sushisource: The point here was to include each failure in the summary at the end of the CI log where people will see it

bergundy: 👍

PR #163, opened 2022-11-17
Add schedule features and support no-workflow config

cretz:
## What was changed

* Added 3 more schedule features
* Moved config loading into Go harness
* Added support for `noWorkflow` bool config in `.config.json`

### DIFF
--- a/features/schedule/trigger/feature.go
+++ b/features/schedule/trigger/feature.go
@@ -1,0 +1,58 @@
+package trigger
+
+import (
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+	"go.temporal.io/sdk-features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+var Feature = harness.Feature{
+	Workflows: Workflow,
+	Execute:   Execute,
+}
+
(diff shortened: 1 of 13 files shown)


features/schedule/trigger/feature.go
```diff
@@ -0,0 +1,58 @@
...
+			r.Log.Warn("Failed deleting schedule handle", "error", err)
+		}
+	}()
+
+	// Trigger it
+	r.Require.NoError(handle.Trigger(ctx, client.ScheduleTriggerOptions{}))
+	// TODO(cretz): We have to wait before triggering again. See
+	// https://github.com/temporalio/temporal/issues/3614
+	time.Sleep(2 * time.Second)
+	r.Require.NoError(handle.Trigger(ctx, client.ScheduleTriggerOptions{}))
```
Quinn-With-Two-Ns, line 49: Are we going to test different overlap options or do we consider that out of scope for this testing?

cretz: Out of scope for just trigger testing. But if we need to test how overlap behaves in a test we can. However there are dozens of fields in schedule we are not testing here in SDK features. I am unsure who is the arbiter of what we should or shouldn't test.

For example, how do we know buffer one works properly with the Go SDK? I can't find an integration test with it.

PR #165, opened 2022-11-17
Add failure converter test

Quinn-With-Two-Ns:
Add a failure converter test to Go, Python, and TS. Java does not have a failure converter API yet. I did not compare all the outputs against a common input like some of the other data converter tests because each SDK has some inconsistencies on how it handles stack traces that are inherited from the language convention/implementation of exceptions.

### DIFF
--- a/features/data_converter/failure/README.md
+++ b/features/data_converter/failure/README.md
@@ -6,6 +6,7 @@ Steps:
 
 - run a workflow that calls an activity that fails with an error and has a `cause` sub error
 - get result payload of ActivityTaskFailed event from workflow history
-- verify the message and potentially stack trace are encoded on the main error and sub error
+- verify the message and stack trace are encoded on the main error and sub error. 
+Failure.message should be "Encoded failure" and Failure.stack_trace should be "". 
 
 Note: Typescript lets us modify the stack trace of the error so we can test more than other languages.
--- a/features/data_converter/failure/README.md
+++ b/features/data_converter/failure/README.md
@@ -1,0 +1,11 @@
+# Failure converter
+
+When using the default failure converter `stack_trace` and `message` can be encoded in the `EncodedAttributes` field.
+
+Steps:
+
+- run a workflow that calls an activity that fails with an error and has a `cause` sub error
+- get result payload of ActivityTaskFailed event from workflow history
+- verify the message and potentially stack trace are encoded on the main error and sub error
+
+Note: Typescript lets us modify the stack trace of the error so we can test more than other languages.
--- a/features/data_converter/failure/failure_converter.ts
+++ b/features/data_converter/failure/failure_converter.ts
@@ -1,4 +1,4 @@
-import { ActivityFailure, DefaultPayloadConverter, DefaultFailureConverter } from '@temporalio/common';
+import { DefaultPayloadConverter, DefaultFailureConverter } from '@temporalio/common';
 
 export const failureConverter = new DefaultFailureConverter({
   payloadConverter: new DefaultPayloadConverter(),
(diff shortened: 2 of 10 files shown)


features/data_converter/failure/failure_converter.ts
```diff
@@ -0,0 +1,6 @@
+import { ActivityFailure, DefaultPayloadConverter, DefaultFailureConverter } from '@temporalio/common';
```
bergundy, line 1: You can put this in `feature.ts`

features/data_converter/failure/README.md
```diff
@@ -0,0 +1,11 @@
+# Failure converter
+
+When using the default failure converter `stack_trace` and `message` can be encoded in the `EncodedAttributes` field.
+
+Steps:
+
+- run a workflow that calls an activity that fails with an error and has a `cause` sub error
```
bergundy, line 7: Mention that you also check that `Failure.stack_trace  == ""` and `Failure.message == "Encoded failure"`

bergundy: Let's bubble the error up from the workflow instead of creating a chain of errors in the activity.

features/data_converter/failure/feature.ts
```diff
@@ -0,0 +1,61 @@
...
+    const cause = failure.cause;
+    assert.ok(cause);
+    assert.equal('Encoded failure', cause.message);
+    assert.equal('', cause.stackTrace);
+    assert.equal('json/plain', cause.encodedAttributes?.metadata?.['encoding']);
+    assert.equal('cause error', JSON.parse(cause.encodedAttributes?.data?.toString() ?? '')['message']);
+    assert.equal('cause stack trace', JSON.parse(cause.encodedAttributes?.data?.toString() ?? '')['stack_trace']);
+  },
+  dataConverter: {
+    failureConverterPath: __dirname + '/failure_converter.js',
```
bergundy, line 59: This won't work on windows but I'd put the converter in the same file anyways.

features/data_converter/failure/feature.go
```diff
@@ -0,0 +1,78 @@
...
+func FailureActivity(ctx context.Context) error {
+	return temporal.NewApplicationErrorWithCause("main error", "customType", errors.New("cause error"))
+}
+
+func CheckResult(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) error {
+	if err := run.Get(ctx, nil); err != nil {
+		return err
+	}
+
+	history := runner.Client.GetWorkflowHistory(ctx, run.GetID(), "", false, 0)
```
cretz, line 50: I think it'd be clearer to just return error out of the workflow and check the workflow error and causes instead of history. It's a clearer end-to-end test. (same for other langs)

Quinn-With-Two-Ns: I'm fine with returning it from the workflow, but If I don't check the history how do I know if the message and stack trace got encoded ?

features/data_converter/failure/README.md
Quinn-With-Two-Ns: done!

features/data_converter/failure/feature.ts
bergundy: Please change to use `path.join` or `require.resolve('./failure_converter')`

bergundy (approved):

features/data_converter/failure/feature.go
cretz: Not sure what I was thinking back when I made this comment. I probably just wasn't paying attention that you assert the attributes.

PR #169, opened 2022-11-23
Test eager activity + non-remote-activities worker

bergundy:
Verifying that the worker doesn't crash with this combination as observed and fixed in Core based SDKs.

### DIFF
--- a/features/eager_activity/non_remote_activities_worker/feature.py
+++ b/features/eager_activity/non_remote_activities_worker/feature.py
@@ -1,0 +1,42 @@
+from datetime import timedelta
+
+from temporalio import activity, workflow
+from temporalio.exceptions import ActivityError, TimeoutError, TimeoutType
+from temporalio.worker import WorkerConfig
+
+from harness.python.feature import register_feature
+
+
+@workflow.defn
+class Workflow:
+    @workflow.run
+    async def run(self) -> None:
+        # Run a workflow that schedules a single activity with short schedule-to-close timeout
+        try:
+            await workflow.execute_activity(
+                dummy,
+                # Pick a long enough timeout for busy CI but not too long to get feedback quickly
+                schedule_to_close_timeout=timedelta(seconds=3),
+            )
+            raise RuntimeError("Expected activity to time out")
+        except ActivityError as e:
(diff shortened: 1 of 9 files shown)


features/eager_activity/non_remote_activities_worker/feature.py
```diff
@@ -0,0 +1,42 @@
...
+                # Pick a long enough timeout for busy CI but not too long to get feedback quickly
+                schedule_to_close_timeout=timedelta(seconds=3),
+            )
+            raise RuntimeError("Expected activity to time out")
+        except ActivityError as e:
+            # Catch activity failure in the workflow, check that it is caused by schedule-to-start timeout
+            if not isinstance(e.cause, TimeoutError):
+                raise e
+
+            assert isinstance(e.cause, TimeoutError)
```
cretz, line 27: Already effectively done by the conditional above

features/eager_activity/non_remote_activities_worker/feature.py
```diff
@@ -0,0 +1,42 @@
...
+@activity.defn
+async def dummy() -> None:
+    pass
+
+
+# Start a worker with activities registered and non-local activities disabled
+register_feature(
+    workflows=[Workflow],
+    activities=[dummy],
+    worker_config=WorkerConfig(no_remote_activities=True),
```
cretz, line 41: Alternatively could have just left off `activities` here when registering, though I understand if this wants to test the core side of polling or not only

cretz (approved):

features/eager_activity/non_remote_activities_worker/feature.py
bergundy: Yeah, this tests the case where the worker supports local activities and no remote activities.

features/eager_activity/non_remote_activities_worker/feature.py
bergundy: I needed this one for mypy to infer the type, can remove the condition above.

bergundy:
Merging even though some of the tests are flaky.
The flakes are not related to my changes.

I'll open an issue to fix the flakes.

Spikhalskiy:
This PR has a stable failing test for JavaSDK: eager_activity/non_remote_activities_worker

PR #171, opened 2022-11-23
Support server triggering sdk-features runs

Sushisource:
Draft while building


.github/workflows/go.yaml
```diff
@@ -30,6 +34,18 @@ jobs:
         run: 'echo head_ref: ${{ github.head_ref }}, ref: ${{ github.ref }}, go sdk version: ${{ inputs.version }}'
         working-directory: '.'
 
+      - name: Download docker artifacts
+        if: ${{ inputs.docker-image-artifact-name }}
+        uses: actions/download-artifact@v3
```
cretz, line 39: Are we expecting to build and upload a docker image for Temporal server on every PR? Ideally we could provide a server _commit/ref_ and build the server at test time. Or can we share the docker container built on the server side with this somehow without having to push then download? Or am I misunderstanding?

Granted I think the same about the constant SDK features containers too. We may be adding unnecessary steps containerizing some of these things.

cmd/run.go
```diff
@@ -191,6 +191,13 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
 		defer server.Stop()
 		r.config.Server = server.FrontendHostPort
 		r.log.Info("Started server", "HostPort", r.config.Server)
+	} else if !strings.Contains(r.config.Server, "tmprl.cloud") {
```
cretz, line 194: This is fickle to rely on string pieces like this. There are other ways to do quick checks on whether namespace is available. I suspect you can change `WaitNamespaceAvailable` to do a never-match list and check for https://pkg.go.dev/go.temporal.io/api/serviceerror#NamespaceNotFound.

What you've found is that we don't have a reasonable server-agnostic way to check if a namespace exists. We should hack one and consider this a pain point for fixing for all users, not just us.

features/schedule/basic/feature.go
```diff
@@ -55,15 +55,20 @@ func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error)
 	r.Require.Equal(workflowID, desc.Schedule.Action.(client.ScheduleWorkflowAction).ID)
 
 	// Confirm simple list
-	iter, err := r.Client.ScheduleClient().List(ctx, client.ScheduleListOptions{})
-	r.Require.NoError(err)
-	foundSchedule := false
-	for iter.HasNext() && !foundSchedule {
-		e, err := iter.Next()
+	// TODO: This should not require a retry loop. See https://github.com/temporalio/sdk-features/issues/182
```
cretz, line 58:
```suggestion
	// Unfortunately better visibility means worse consistency in some environments. See https://github.com/temporalio/sdk-features/issues/182
```

features/schedule/basic/feature.go
```diff
@@ -55,15 +55,20 @@ func Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error)
...
+	listingErr := harness.RetryFor(10, 1*time.Second, func() (bool, error) {
```
cretz, line 59: Alternatively, `r.Require.Eventually(...`, see https://pkg.go.dev/github.com/stretchr/testify/require#Eventually

harness/go/harness/util.go
```diff
@@ -29,3 +31,46 @@ func FindEvent(history client.HistoryEventIterator, cond func(*historypb.History
 	}
 	return nil, nil
 }
+
+func WaitNamespaceAvailable(ctx context.Context, hostPortStr string, namespace string) error {
```
cretz, line 35: Add Godoc to exported functions

.github/workflows/go.yaml
Sushisource: You're misunderstanding. The server triggered workflows build a docker image which is used internally as an artifact. Nothing is pushed.

Building the docker image takes essentially an equivalent amount of time to just building the server, but makes using the built artifacts dramatically easier across parallel runs.

cmd/run.go
Sushisource: Sure, but we've done the other way in tests before and it's got it's own downsides. I agree this should be easier but I don't really think it's in the scope of this PR

cretz: > I don't really think it's in the scope of this PR

Agree about the general lack of namespace-exists, but this PR adds a brand new `WaitNamespaceAvailable`. Can't we just make that work in all environments? (via hacks or otherwise)

.github/workflows/java.yaml
```diff
@@ -58,5 +74,13 @@ jobs:
...
+        if: ${{ inputs.docker-image-artifact-name }}
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml up -d temporal-server cassandra elasticsearch
+
       - name: Run SDK-features tests
-        run: go run . run --lang java --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-java)' || inputs.version }}"
+        run: go run . run --lang java ${{ inputs.docker-image-artifact-name && '--server localhost:7233 --namespace default' || ''}} --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-java)' || inputs.version }}"
+
+      - name: Tear down docker compose
+        if: inputs.docker-image-artifact-name && (success() || failure())
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml down -v
```
mindaugasrukas, line 86: steps looks duplicated

.github/workflows/python.yaml
```diff
@@ -73,8 +89,16 @@ jobs:
...
       - name: Build sdk-features go
         run: go build
 
       - name: Run SDK-features tests
-        run: go run . run --lang py --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-python)' || inputs.version }}"
+        run: go run . run --lang python ${{ inputs.docker-image-artifact-name && '--server localhost:7233 --namespace default' || ''}} --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-python)' || inputs.version }}"
+
+      - name: Tear down docker compose
+        if: inputs.docker-image-artifact-name && (success() || failure())
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml down -v
```
mindaugasrukas, line 104: steps looks duplicated

.github/workflows/typescript.yaml
```diff
@@ -78,5 +94,13 @@ jobs:
...
+        if: ${{ inputs.docker-image-artifact-name }}
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml up -d temporal-server cassandra elasticsearch
+
       - name: Run SDK-features tests
-        run: go run . run --lang ts --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-ts)' || inputs.version }}"
+        run: go run . run --lang ts ${{ inputs.docker-image-artifact-name && '--server localhost:7233 --namespace default' || ''}} --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-ts)' || inputs.version }}"
+
+      - name: Tear down docker compose
+        if: inputs.docker-image-artifact-name && (success() || failure())
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml down -v
```
mindaugasrukas, line 106: steps looks duplicated

cmd/github.go
```diff
@@ -0,0 +1,35 @@
...
+const errFileCmdFmt = "failed to write to github file: %v"
+
+// Set a GitHub environment value. Only works with values without a linebreak.
+func writeGitHubEnv(name string, value string) (retErr error) {
+	filepath := os.Getenv("GITHUB_ENV")
+	if filepath == "" {
+		// Just don't do anything if we're not running in a GH env
+		return nil
+	}
+	msg := []byte(fmt.Sprintf("%s=%s", name, value))
```
mindaugasrukas, line 17: nit: can we move `msg` close to the usage for better readability?

.github/workflows/go.yaml
```diff
@@ -52,5 +68,13 @@ jobs:
...
+        if: ${{ inputs.docker-image-artifact-name }}
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml up -d temporal-server cassandra elasticsearch
+
       - name: Run SDK-features tests
-        run: go run . run --lang go --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-go)' || inputs.version }}"
+        run: go run . run --lang go ${{ inputs.docker-image-artifact-name && '--server localhost:7233 --namespace default' || ''}} --version "${{ inputs.version-is-repo-ref && '$(realpath ../sdk-go)' || inputs.version }}"
+
+      - name: Tear down docker compose
+        if: inputs.docker-image-artifact-name && (success() || failure())
+        run: docker-compose -f /tmp/server-docker/docker-compose.yml -f ./dockerfiles/docker-compose.for-server-image.yaml down -v
```
mindaugasrukas, line 80: My opinion about CI workflows is that they should be reproducible locally (for development and debugging) and should be documented for contributors.

It's hard to maintain (keep them in sync) three separate parts: documentation, CI workflow steps, and local development workflow. I suggest having a single command via build manager `make`, `mage`, or similar. And everywhere have a simple command such as `make start-server-and-deps; make test; make teardown`. And then have a single place for maintaining them.

.github/workflows/ci.yaml
```diff
@@ -110,46 +110,13 @@ jobs:
...
-      # Latest release at the time this was written
-      # TODO(bergundy): Find some way to automatically upgrade to "latest"
-      sdk-version: 'v0.1b3'
-      semver-latest: none
-      run-tag: '0.1b3'
+      do-push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+      go-ver: 'v1.18.1'
+      ts-ver: 'v1.4.4'
+      java-ver: 'v1.17.0'
+      py-ver: 'v0.1b3'
```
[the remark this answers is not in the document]
cretz: We manually update this file when we do an SDK release (these versions are in other spots in this file and these versions are the ones included at the root of this repo for local development)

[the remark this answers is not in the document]
Sushisource: There's a TODO above about possibly automating this later. It's not the most straightforward thing and I think outside the scope of this first PR.

cmd/run.go
Sushisource: I'll just do what we did in TS where you try to describe a nonexistent workflow make sure it says it can't find the workflow rather than the namespace not existing. https://github.com/temporalio/sdk-typescript/blob/e283b893d05ef4777e8ce277413e542ce69ef334/scripts/wait-on-temporal.mjs#L20

Annoying, but, whatever.

features/schedule/basic/feature.go
Sushisource: Nice to know

.github/workflows/go.yaml
Sushisource: I don't want to introduce `make` since we haven't needed it yet. That said I'd be fine with adding a few little bash scripts into the `.github` directory that could be used for this purpose.

The commands cannot be combined into one line without breaking the `if` guards, which are important for visibility purposes in the CI UI. I wouldn't want someone to need to walk the script just to understand why something is or is not happening.

That sound good?

cretz: I would strongly discourage `make` as it is not a good cross platform tool. We have a CLI right here called sdk-features. If we have to move this simple one line to inside of the CLI, ok. You're not gaining much though.

Sushisource: Given these bits are only intended to run on CI, I'd just take the ease of a shell script here

cretz: > Given these bits are only intended to run on CI

Conflicts with

> My opinion about CI workflows is that they should be reproducible locally

If they're only for CI, they can be in the CI language of choice (YAML). If the requirement is "reproducible locally" we should use this repo's language of choice (Go) and not something platform specific.

Sushisource: Well, because this
> Why is docker-compose + args not ok but make + args ok?

Was basically my thought. Though we can change it to "deps up" and "deps down" or something instead of passing the docker files & args. I agree it's not much savings. I mean, I can just move it into Go and be done with it but it does feel a little overkill. @mindaugasrukas are you fine with that?

.github/workflows/docker-images.yaml
```diff
@@ -66,17 +66,28 @@ jobs:
...
+        if: inputs.do-push
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Push image to DockerHub
+        if: inputs.do-push
+        # TODO: Push all tags that we built
+        run: |
+          docker tag ${{ env.SDK_FEAT_BUILT_IMAGE_TAG }} temporaliotest/sdk-features:${{ inputs.lang }}
```
Sushisource, line 92: I need to get this merged to be able to test it unfortunately. You can't run manually triggered GH workflows that aren't in the default branch, although, amusingly, once they are you can point at a different branch's implementation to use. So I'll finish this step in another PR.

cretz (approved):

.github/workflows/go.yaml
mindaugasrukas: `make` was mentioned here as an example. Bash or other scripts is fine.

My point here is it's hard to keep documentation in sync with CI workflow and development workflow definitions. Also, having CI workflow execution steps defined in a YAML makes it complicated to lint, test, and reproduce them locally. Yes, it's only a few lines of code today, but there will be many more tomorrow. And then reproduction locally becomes many copy-pasting with potential errors.

An alternative (which works for me) is to call a script inside workflows definitions: I.e., `run: bash setup-something.sh <params...>`, `run: bash run-test.sh <params>` and similar, and move all the logic to those scripts. That makes it easy to keep all workflows in sync, lint command, test logic, and reproduce CI runs locally. All parameters might be defined as env variables. So documentation and reproduction steps locally might look like this:
```
export PARAM_1=value-1
export PARAM_2=value-1
export PARAM_3=value-1
...
bash setup-something.sh
bash run-test.sh
bash cleanup.sh
...
```

This is a proposal from my personal experience. If that something is not applicable here, it's fine to keep it as is.

.github/workflows/docker-images.yaml
mindaugasrukas: Good to know.
I guess it's enough to create a dummy workflow on the main branch and test it against the development branch.

.github/workflows/go.yaml
mindaugasrukas: Another argument for scripts is that there are four duplicated commands in this PR, and they all must be in sync. Having scripts makes them easy to maintain.

mindaugasrukas (approved):

.github/workflows/go.yaml
Sushisource: Yeah, I absolutely agree with the philosophy here, I just think we're not quite at that point with the CI definitions. They're all one liners, and the way you'd run the docker compose stuff specifically locally is necessarily different from in CI since in CI the images are downloaded from the last  run's artifacts, which you can't do locally. If they were more than one-line invocations then yeah I think we'll want to move them inside the Go runner.

Then the invocation of the actual test is just a normal one-liner you'd want to control yourself locally.

As for the repetition across the definitions per language, yeah, that's a bit annoying. If/when these get more complex, I think solving that by templating them is probably the most straightforward way.

PR #176, opened 2022-11-30
Checkout this repo properly on PRs

Sushisource:
This repo wasn't checking out the PR's changes when re-using the features workflows

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -12,6 +12,12 @@ on:
       version-is-repo-ref:
         required: true
         type: boolean
+      sdk-features-repo-path:
+        type: string
+        default: 'temporalio/sdk-features'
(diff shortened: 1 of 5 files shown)


.github/workflows/go.yaml
```diff
@@ -12,6 +12,12 @@ on:
       version-is-repo-ref:
         required: true
         type: boolean
+      sdk-features-repo-path:
```
cretz, line 15: These vars need to be set to be used below, correct? (same for other ones of these that aren't using these vars)

bergundy (approved):

.github/workflows/go.yaml
Sushisource: Derp. Thanks.

cretz (approved):

PR #185, opened 2022-12-07
Update Python SDK to 0.1b4

### LINKED ISSUE #184: [Bug] Python features/query/timeout_due_to_no_active_workers should check for deadline exceeded
### Describe the bug

Other langs check for the right error type, this one does not because it was left over from manual client-side timeout

cretz:
## What was changed

* Update Python SDK to 0.1b4
* Fix race from #184

## Checklist

1. Closes #184

### DIFF
--- a/features/query/timeout_due_to_no_active_workers/feature.py
+++ b/features/query/timeout_due_to_no_active_workers/feature.py
@@ -34,8 +34,12 @@ async def check_result(runner: Runner, handle: WorkflowHandle):
     try:
         await handle.query(Workflow.simple_query, rpc_timeout=timedelta(seconds=1))
     except RPCError as e:
-        # Cancelled rather than deadline exceeded since the timeout is client-side
-        assert e.status == RPCStatusCode.CANCELLED
+        # Can be cancelled or deadline exceeded depending on whether client or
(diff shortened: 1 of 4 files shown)


features/query/timeout_due_to_no_active_workers/feature.py
```diff
@@ -34,8 +34,12 @@ async def check_result(runner: Runner, handle: WorkflowHandle):
     try:
         await handle.query(Workflow.simple_query, rpc_timeout=timedelta(seconds=1))
     except RPCError as e:
-        # Cancelled rather than deadline exceeded since the timeout is client-side
-        assert e.status == RPCStatusCode.CANCELLED
+        # Can be cancelled or deadline exceeded depending on whether client or
+        # server hit timeout first in a racy way
```
Quinn-With-Two-Ns, line 38: Is this just a temporary work around?

cretz: I cannot replicate deadline exceeded against a docker container or Temporalite that happened in #184. Given my inability to replicate, I have to assume that this is the server racing against core gRPC layer (tonic) to see which hits the deadline first.

I'd accept alternative suggestions here.

PR #189, opened 2022-12-09
Enable the pushing of all image tags

Sushisource:
Ensure all image tags get published when publishing is on

### DIFF
--- a/cmd/push_images.go
+++ b/cmd/push_images.go
@@ -1,0 +1,70 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/urfave/cli/v2"
+	"os"
+	"os/exec"
+	"strings"
+)
+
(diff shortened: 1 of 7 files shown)


cmd/push_images.go
```diff
@@ -0,0 +1,70 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/urfave/cli/v2"
```
cretz, line 5: May need to run "go fmt" or one w/ imports (I forget which, my IDE does it) do separate stdlib imports from third parties.

cretz (approved):

cmd/push_images.go
Sushisource: I'm using Goland set to do it, `go fmt` changes nothing here either. It looks like you have to tick an explicit setting in Goland for that, which I've done now

PR #192, opened 2022-12-19
Add schedule feature test

Quinn-With-Two-Ns:
Add schedule feature tests to mirror the go tests

### DIFF
--- a/features/schedule/backfill/feature.ts
+++ b/features/schedule/backfill/feature.ts
@@ -57,8 +57,6 @@ export const feature = new Feature({
           });
         }, 10)
       );
-
-      return undefined;
     } finally {
       await handle.delete();
     }
--- a/features/schedule/backfill/feature.ts
+++ b/features/schedule/backfill/feature.ts
@@ -1,0 +1,66 @@
+import { Feature, retry } from '@temporalio/harness';
+import { ScheduleClient, Connection, ScheduleOverlapPolicy } from '@temporalio/client';
+import * as assert from 'assert';
+import { randomUUID } from 'crypto';
+
+export async function workflow(arg: string): Promise<string> {
+  return arg;
+}
+
+export const feature = new Feature({
+  workflow,
+  alternateRun: async (runner) => {
+    const connection = await Connection.connect({
+      address: runner.options.address,
+      tls: runner.options.tlsConfig,
+    });
+
+    const client = new ScheduleClient({
+      connection,
+      namespace: runner.options.namespace,
+    });
+    const scheduleId = `schedule-backfill-${randomUUID()}`;
+    const handle = await client.create({
+      scheduleId,
+      spec: {
+        intervals: [{ every: '1m' }],
+      },
+      action: {
+        type: 'startWorkflow',
+        workflowType: workflow,
+        taskQueue: runner.options.taskQueue,
+      },
+    });
+
+    try {
+      const now = new Date();
+      const threeYearsAgo = new Date(new Date(now).setFullYear(now.getFullYear() - 3));
+      const thirtyMinutesAgo = new Date(new Date(now).setMinutes(now.getMinutes() - 30));
+
+      await handle.backfill([
+        {
+          start: new Date(new Date(threeYearsAgo).setMinutes(threeYearsAgo.getMinutes() - 2)),
+          end: threeYearsAgo,
+          overlap: ScheduleOverlapPolicy.ALLOW_ALL,
+        },
+        {
+          start: new Date(new Date(thirtyMinutesAgo).setMinutes(thirtyMinutesAgo.getMinutes() - 2)),
+          end: thirtyMinutesAgo,
+          overlap: ScheduleOverlapPolicy.ALLOW_ALL,
+        },
+      ]);
+
+      assert.ok(
+        await retry(async function () {
+          return handle.describe().then((s) => {
+            return s.info.numActionsTaken == 4;
+          });
+        }, 10)
+      );
+
+      return undefined;
+    } finally {
+      await handle.delete();
+    }
+  },
+});
--- a/features/schedule/basic/feature.ts
+++ b/features/schedule/basic/feature.ts
@@ -80,9 +80,7 @@ export const feature = new Feature({
       };
 
       // Wait for first completion
-      assert.ok(
-        await retry( () => waitCompletedWith('arg1'), 10)
-      );
+      assert.ok(await retry(() => waitCompletedWith('arg1'), 10));
 
       // Update
       await handle.update((x) => ({
--- a/features/schedule/basic/feature.ts
+++ b/features/schedule/basic/feature.ts
@@ -1,0 +1,112 @@
+import { Feature, retry } from '@temporalio/harness';
+import { ScheduleClient, Connection, ScheduleOverlapPolicy } from '@temporalio/client';
+import * as assert from 'assert';
+import { randomUUID } from 'crypto';
+
+export async function BasicWorkflow(arg: string): Promise<string> {
+  return arg;
+}
+
+export const feature = new Feature({
+  workflow: BasicWorkflow,
+  alternateRun: async (runner) => {
+    const connection = await Connection.connect({
+      address: runner.options.address,
+      tls: runner.options.tlsConfig,
+    });
+
+    const client = new ScheduleClient({
+      connection,
+      namespace: runner.options.namespace,
+    });
+    const scheduleId = `schedule-basic-${randomUUID()}`;
+    const workflowId = `schedule-basic-workflow-${randomUUID()}`;
+    const handle = await client.create({
+      scheduleId,
+      spec: {
+        intervals: [{ every: '2s' }],
+      },
+      action: {
+        type: 'startWorkflow',
+        workflowId,
+        workflowType: BasicWorkflow,
+        args: ['arg1'],
+        taskQueue: runner.options.taskQueue,
+      },
+      policies: {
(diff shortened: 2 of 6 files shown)


Quinn-With-Two-Ns:
CI is failing on listing schedules, the server is not returning any schedules. Seems to only be on temporalite.
```
2022-12-19T23:06:10.766Z [INFO] Worker state changed { state: 'STOPPED' }
Error: 1 feature(s) failed: 
Feature schedule/basic failed with AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert.ok(await (0, harness_1.retry)(async function () {
      for await (const schedule of client.list({ pageSize: 1 })) {
          if (schedule.scheduleId == scheduleId) {
              return true;
          }
      }
      return false;
  }, 10))
```
@mjameswh  have you ever seen this before? I'd think it was a server bug, but it works in the go SDK

features/schedule/backfill/feature.ts
```diff
@@ -0,0 +1,66 @@
...
+        type: 'startWorkflow',
+        workflowType: workflow,
+        taskQueue: runner.options.taskQueue,
+      },
+    });
+
+    try {
+      const now = new Date();
+      const threeYearsAgo = new Date(new Date(now).setFullYear(now.getFullYear() - 3));
+      const thirtyMinutesAgo = new Date(new Date(now).setMinutes(now.getMinutes() - 30));
```
bergundy, line 38: You need the wrapping `Date` object, `Date` is mutable (unfortunately)

features/schedule/backfill/feature.ts
```diff
@@ -0,0 +1,66 @@
+import { Feature, retry } from '@temporalio/harness';
+import { ScheduleClient, Connection, ScheduleOverlapPolicy } from '@temporalio/client';
+import * as assert from 'assert';
+import { randomUUID } from 'crypto';
+
+export async function workflow(arg: string): Promise<string> {
+  return arg;
```
bergundy, line 7: Is there any information provided to the workflow from the scheduler letting it know what period it's being backfilled for?
That's super useful in systems like airflow for example.
We should add that feature if it doesn't exist and test it if it does.

features/schedule/backfill/feature.ts
bergundy: OIC the setters don't return a date :(
Wow this Date API sucks.

features/schedule/basic/feature.ts
```diff
@@ -0,0 +1,112 @@
...
+      );
+
+      const waitCompletedWith = async (untilResult: string) => {
+        const iterable = await runner.client.list({
+          query: "WorkflowType = 'BasicWorkflow'",
+        });
+        for await (const exec of iterable) {
+          if (!exec.workflowId.startsWith(workflowId)) {
+            continue;
+          }
```
bergundy, line 67: Is this additional filtering required due to limitations of "basic visibility"?
Can we add a TODO to fix the query once this is supported?

features/schedule/basic/feature.ts
```diff
@@ -0,0 +1,112 @@
...
+          if (exec.status.name == 'COMPLETED') {
+            const wfHandle = runner.client.getHandle(exec.workflowId, exec.runId);
+            const result = await wfHandle.result();
+            if (result == untilResult) {
```
bergundy, line 71: nit: use `===`

features/schedule/basic/feature.ts
```diff
@@ -0,0 +1,112 @@
...
+          }
+        }
+        return false;
+      };
+
+      // Wait for first completion
+      assert.ok(
+        await retry(async function () {
+          return waitCompletedWith('arg1');
+        }, 10)
```
bergundy, line 85: nit:

```suggestion
        await retry(() => waitCompletedWith('arg1'), 10)
```

features/schedule/basic/feature.ts
```diff
@@ -0,0 +1,112 @@
...
+      }));
+
+      // Wait for second completion
+      assert.ok(
+        await retry(async function () {
+          return waitCompletedWith('arg2');
+        }, 10)
+      );
+
+      return undefined;
```
bergundy, line 107: That's implicit, no need for this return statement.

features/schedule/pause/feature.ts
```diff
@@ -0,0 +1,66 @@
...
+      await handle.unpause();
+      await assertState(false, 'Unpaused via TypeScript SDK"');
+      // Re-unpause
+      await handle.unpause('custom note2');
+      await assertState(false, 'custom note2');
+      // Pause
+      await handle.pause();
+      await assertState(true, 'Paused via TypeScript SDK"');
+
+      return undefined;
```
bergundy, line 61: Same here, no need for this explicit return

features/schedule/trigger/feature.ts
```diff
@@ -0,0 +1,63 @@
+import { Feature, retry } from '@temporalio/harness';
+import { ScheduleClient, Connection } from '@temporalio/client';
+import * as assert from 'assert';
+import { randomUUID } from 'crypto';
+
+function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
bergundy, line 8: Can use this API from node 16: https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options

features/schedule/trigger/feature.ts
```diff
@@ -0,0 +1,63 @@
...
+      await delay(2000);
+
+      assert.ok(
+        await retry(async function () {
+          return handle.describe().then((s) => {
+            return s.info.numActionsTaken === 2;
+          });
+        })
+      );
+      return undefined;
```
bergundy, line 58: same here

harness/ts/harness.ts
```diff
@@ -286,3 +286,18 @@ export class Runner<W extends Workflow, A extends UntypedActivities> {
     return history;
   }
 }
+
+export async function sleep(ms: number): Promise<void> {
+  return new Promise((r) => setTimeout(r, ms));
+}
```
bergundy, line 292: Same here, can use built-in promise based timers.

harness/ts/harness.ts
```diff
@@ -286,3 +286,18 @@ export class Runner<W extends Workflow, A extends UntypedActivities> {
...
+
+export async function retry(fn: () => Promise<boolean>, retries = 3, duration = 1000): Promise<boolean> {
+  if (!retries) {
+    return false;
+  }
+  if (!(await fn())) {
+    await sleep(duration);
+    return retry(fn, retries - 1, duration);
```
bergundy, line 300: Better to use a loop than recursion here, JS has a pretty low default limit of the call stack.

features/schedule/backfill/feature.ts
Quinn-With-Two-Ns: I am not super familiar with airflow, what information does it expose? The only additional information a workflow has when started be a schedule is :

```
Workflow Executions started by a Schedule have the following additional properties:

* The Action's timestamp is appended to the Workflow Id.

* The TemporalScheduledStartTime [Search Attribute](https://docs.temporal.io/concepts/what-is-a-search-attribute) is added to the Workflow Execution. 

* The value is the Action's timestamp.
The TemporalScheduledById Search Attribute is added to the Workflow Execution. The value is the Schedule Id.
```

bergundy: That's great, can we test this?

Quinn-With-Two-Ns: I don't think those properties listed expose any backfill information. It should definitely be tested, but does that belong here? That seems like a server test irrelevant of any sdk ?

features/schedule/basic/feature.ts
Quinn-With-Two-Ns: I don't think it is a limitation of basic visibility? We want to match just on the beginning and since `WorkflowID` is a keyword it only supports full matching regardless of basic vs advanced visibility.

features/schedule/backfill/feature.ts
bergundy: We're testing Temporal features as a whole here, this repo should have just been named `features`. SDKS are just a part of what makes up a feature.

features/schedule/basic/feature.ts
bergundy: OIC, I missed that we're using `startsWith`, is there another search attribute we can leverage?

mjameswh: I'm facing a similar issue in another test. On most SQL backends, startwith x can be implemented using a Between x and xzzzzzz... but that idiom seems to fail in SQLite...

mjameswh: I considered generating a random task queue name, but that attribute is not searchable in sql backends. Also considered to make the workflow type random, but that'a ridiculously complex to implement because we don't have catch all workflows handlers...

Quinn-With-Two-Ns: Part of what this is testing is that the workflow ID has the correct form so there is no other search attribute that can help us do that.

features/schedule/backfill/feature.ts
Quinn-With-Two-Ns: Hm OK I was under the impression we were using sdk-features for just cross sdk testing. I'll make an issue to test temporal sets the right data in workflow execution.

Quinn-With-Two-Ns: https://github.com/temporalio/sdk-features/issues/202

Quinn-With-Two-Ns:
I'll merge this evening if there are no blockers

PR #217, opened 2023-01-25
Add skipping mechanism and eager workflow features

bergundy:
This isn't ready for review yet.
I'll undraft it once the Go SDK supports the relevant changes so we can run the eager workflow feature tests.


bergundy:
I ended up submitting these as server integration tests since that's a better place for them.
Closing the PR but wondering if we want to extract the commit that allows skipping tests.

@mmcshane @cretz

PR #224, opened 2023-02-03
Update sdk-go dependency to v1.21.0

mmcshane:
## What was changed

Update sdk-go dependency to v1.21.0

## Why?

Good dependency hygiene

## Checklist

1. Closes 

2. How was this tested:

Ran all Go feature tests

3. Any docs updates needed?

No


Quinn-With-Two-Ns:
I think you need to update the `go-ver` in the docker image task

mmcshane:
@Quinn-With-Two-Ns it looks like that's for the Go version not the SDK version

mmcshane:
@Quinn-With-Two-Ns Weird that it was on 1.19? I thought it was for the Go language version. Anyway, updated to 1.21

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
@mmcshane Since I did the 1.20 release you can blame me

PR #230, opened 2023-02-21
Scrub WorkflowTaskStartedEventAttributes.HistorySizeBytes

bergundy:
History comparison is failing now that server populates this field: https://github.com/temporalio/temporal/actions/runs/4210199925/jobs/7315310825#step:10:474

### DIFF
--- a/harness/go/history/history.go
+++ b/harness/go/history/history.go
@@ -183,6 +183,7 @@ func scrubRunSpecificScalars(v interface{}) {
 	case *history.WorkflowTaskStartedEventAttributes:
 		v.Identity = ""
 		v.RequestId = ""
+		v.HistorySizeBytes = 0
 	case *history.WorkflowTaskCompletedEventAttributes:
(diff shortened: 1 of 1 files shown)


harness/go/history/history.go
```diff
@@ -183,6 +183,7 @@ func scrubRunSpecificScalars(v interface{}) {
 	case *history.WorkflowTaskStartedEventAttributes:
 		v.Identity = ""
 		v.RequestId = ""
+		v.HistorySizeBytes = 0
```
rodrigozhou, line 186: Does this fix for both Java and Go?

rodrigozhou (approved):

harness/go/history/history.go
bergundy: It disables checking this field, should fix any SDKs that support comparing histories: Go and Java today.

cretz (approved):

PR #250, opened 2023-03-14
Feature tests for workflow update

mmcshane:
## What was changed

Adds feature tests for the workflow update feature

## Why?

We would like to know if workflow update works.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


features/update/updateutil/shutdown.go
```diff
@@ -0,0 +1,21 @@
...
+
+func AwaitShutdown(ctx workflow.Context) {
+	_ = workflow.GetSignalChannel(ctx, shutdown_signal).Receive(ctx, nil)
+}
+
+func RequestShutdown(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) {
+	runner.Require.NoError(
+		runner.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), shutdown_signal, nil),
+	)
+}
```
Sushisource, line 21: These don't really seem update-specific, no?

Also I had added some support for stopping/restarting workers back in this PR: https://github.com/temporalio/features/pull/112/files#diff-2c10558709e9846fd29a7ee4467a27536755447429e9cb5b5d305c997cdb0953

Maybe you can re-use that?

mmcshane: I had previously been doing something fancier here but now there's a good argument that these functions aren't useful enough to defend their own existence. They're basically saving zero lines of code at the cost of the indirection into another file.

I hadn't noticed the Start/Stop work, thanks for the pointer I'll integrate that.

mmcshane:
Going to need to do the "summary" work so that the harness knows which tests are skipped - the history handling attempting and failing to execute on skipped tests is breaking the harness.

PR #251, opened 2023-03-16
Feature test outcome summarization for Go

mmcshane:
## What was changed

Adds a tcp listener to the features CLI that accepts JSONL data describing feature test results. The JSON objects are expected to have at least three string fields: `name` (test name), `outcome` (PASSED|FAILED|SKIPPED), and `message` (free text). The Go test runner is also updated to emit this data. We use a tcp socket here to support execution under Windows.

## Why?

The overall goal here is to avoid trying to run history replays for tests that have been skipped (e.g. due to a lack of server support for the feature as is currently the case for older servers and the update feature).

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -1,10 +1,14 @@
 package cmd
 
 import (
+	"bufio"
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
+	"net"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -24,6 +28,11 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
+const (
+	summaryListenAddr = "127.7.2.1:0"
+	FeaturePassed     = "PASSED"
+)
+
 func runCmd() *cli.Command {
 	var config RunConfig
 	return &cli.Command{
@@ -36,6 +45,14 @@ func runCmd() *cli.Command {
 	}
 }
 
+type SummaryEntry struct {
+	Name    string `json:"name"`
+	Outcome string `json:"outcome"`
+	Message string `json:"message"`
+}
+
+type Summary []SummaryEntry
+
 // RunConfig is configuration for NewRunner.
 type RunConfig struct {
 	PrepareConfig
@@ -46,6 +63,7 @@ type RunConfig struct {
 	GenerateHistory     bool
 	DisableHistoryCheck bool
 	RetainTempDir       bool
+	SummaryURI          string
 }
 
 // dockerRunFlags are a subset of flags that apply when running in a docker container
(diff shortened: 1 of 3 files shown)


cmd/run.go
```diff
@@ -24,6 +28,11 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
+const (
+	summaryListenAddr = "127.7.2.1:0"
+	FeaturePassed     = "PASSED"
```
Quinn-With-Two-Ns, line 33: Do we need to define this twice?

cmd/run.go
```diff
@@ -24,6 +28,11 @@ import (
 	"go.temporal.io/sdk/log"
 )
 
+const (
+	summaryListenAddr = "127.7.2.1:0"
```
Quinn-With-Two-Ns, line 32: any reason we specify "127.7.2.1" over "localhost"?

mmcshane: Down with the 127.0.0.1 hegemony!

Sushisource: I didn't even realize you could bind to 127.xxx things other than .0.0.1. Wild.

cmd/run.go
mmcshane: I thought about introducing a separate module defining the protocol that both the harness and the cmd modules could depend on but it seemed like a too much complexity just to share some constants. Absent that, it seems reasonable for both the producer (the lang-specific runner) and the consumer (the CLI driver) to have this constant. Java, TS, Python, etc. will have it eventually as well.

cmd/run.go
mmcshane: It's a whole /8! 16 million addresses and yet everything wants to fight for 127.0.0.1:8080.

harness/go/cmd/run.go
```diff
@@ -146,28 +187,54 @@ func (r *Runner) Run(ctx context.Context, run *Run) error {
...
-		} else if feature.SkipReason != "" {
-			r.log.Warn("Skipping feature", "Feature", feature.Dir, "Reason", feature.SkipReason)
-			continue
 		}
+		err := func() error {
+			sumEntry := struct {
+				Name    string `json:"name"`
+				Outcome string `json:"outcome"`
+				Message string `json:"message"`
+			}{
```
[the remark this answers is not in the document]
mmcshane: It's just only needed it in the one function on the producer side.

Sushisource (approved):
Makes sense to me overall. Just the one bit about the unused summary entry struct.

harness/go/cmd/run.go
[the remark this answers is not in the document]
Sushisource: You've got a version of him defined up above too though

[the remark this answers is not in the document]
Sushisource: https://github.com/temporalio/features/pull/251/files/8bca7ca6c34ad8421a574568e01e8aba227385dc#diff-a14ccf0f987e243cdab955e7f6087f2fd1e10e84ef3e1e6ef41b9e9565698be2R131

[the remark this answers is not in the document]
mmcshane: Oh shoot. Good catch, thank you. That's a leftover.

cmd/run.go
dandavison: Hi @mmcshane, on a fresh laptop (MacOS Ventura 13.4.1) I'm finding that I need to run

```
sudo ifconfig lo0 alias 127.7.2.1
```

otherwise I get

```
listen tcp 127.7.2.1:0: bind: can't assign requested address
```

Just mentioning in case you know why that IP isn't available as a loopback by default / whether we need to do something to stop this being an issue for others newcomers in the future.

PR #252, opened 2023-03-16
Feature tests for workflow update

mmcshane:
## What was changed

Add feature tests for workflow update

## Why?

Check that e2e integration and some edge cases are functional for workflow update feature.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


mmcshane:
FYI these have been stable for a while now but needed the test harness to be able to skip tests without erroring out.

features/update/activities/feature.go
```diff
@@ -0,0 +1,85 @@
...
+			updateActivity,
+		)
+		runner.Require.NoError(err)
+
+		var result int
+		runner.Require.NoError(handle.Get(ctx, &result))
+		runner.Require.Equal(activityResult*activityCount, result)
+
+		runner.Require.NoError(runner.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), shutdownSignal, nil))
+		updateutil.RequireNoUpdateRejectedEvents(ctx, runner)
```
Quinn-With-Two-Ns, line 46: Do we need this check here since `UpdateHandlerOptions` doesn't have a validator? Or is the intention to have this check for every sync. update test? I notice it was missing from the panic test case?

features/update/validation_replay/README.md
```diff
@@ -0,0 +1,11 @@
+# Replay does not perform update validation
```
Quinn-With-Two-Ns, line 1: Should we also test the update function does run on replay?

features/update/non_durable_reject/README.md
```diff
@@ -0,0 +1,12 @@
+# Update Rejections Do Not Take Up Space In History
+
+When updates are registered with a validation handler and that validation
+handler rejects an update, neither the invocation of the update nor the
```
Quinn-With-Two-Ns, line 4: > nor the rejection nor the rejection

features/update/self/README.md
```diff
@@ -0,0 +1,10 @@
+# Workflow sends an update to itself
+
+Workflows can send updates back to themselves
+
+# Detailed spec
+
+A workflow can use the workflowservice client in an activity to send an update
```
Quinn-With-Two-Ns, line 7: More a question on updates use case: In this test if there was some failure your activity to send a update could be called more then one correct? Could update validators be used to dedup the update if a UUID was attached?

Quinn-With-Two-Ns (commented):
Do we ever test am update returning in error in the update handler function?

features/update/activities/feature.go
mmcshane: I think my intent was to add it to all of them. That it's missing from the panic test is just an oversight - I'll add it.

features/update/validation_replay/README.md
mmcshane: Maybe I'm wrong about what the harness does during replay but if the update didn't run under replay then we wouldn't see the activity and I would have expected that to get detected as mismatched history?

features/update/non_durable_reject/README.md
mmcshane: :+1:

features/update/self/README.md
mmcshane: We made the choice not to include the UpdateID in the invocation of the handler but the user is certainly free to include an idempotency token in their application-layer update argument. The server can and does dedup by update ID in that a second request with the same update ID will wait on the same update protocol instance in the history service. You can think of this as finding and waiting on the same Future-ish Get()

features/update/validation_replay/README.md
Quinn-With-Two-Ns: No your right ignore me. Yeah the history should fail determinism if the activity didn't run

Quinn-With-Two-Ns (approved):
LGTM, though I am not sure how much coverage we want with these tests? 

Two tests we don't have now that I think would be useful are testing updates with queries and testing update handlers that return an error.

PR #259, opened 2023-04-12
Update go sdk to v1.22.0

Quinn-With-Two-Ns:
Update Go SDK to v1.22.0


cretz:
Might as well wait for 1.22.1

PR #269, opened 2023-04-27
Refactor to external SDK builder library

cretz:
## What was changed

Created a new package called `sdkbuild` that has a slight abstraction to build binaries that include SDK dependencies at arbitrary versions or local clone paths. This is needed so other projects can make SDK-based builds too.


sdkbuild/typescript.go
```diff
@@ -0,0 +1,201 @@
...
+		if _, err := os.Stat(filepath.Join(options.Version, "package.json")); err != nil {
+			return nil, fmt.Errorf("failed finding package.json in version dir: %w", err)
+		}
+		localPath := "file:" + options.Version
+		pkgs := []string{"activity", "client", "common", "internal-workflow-common",
+			"internal-non-workflow-common", "proto", "worker", "workflow"}
+		for _, pkg := range pkgs {
+			packageJSONDepStr += fmt.Sprintf(`"@temporalio/%v": "%v/packages/%v",`, pkg, localPath, pkg)
+			packageJSONDepStr += "\n    "
+		}
```
Sushisource, line 83: I really think we should keep the template. It's eminently more readable than this (and already was set up). This feels like a step backwards.

The others probably deserve to be as well, but we don't need to do that right now.

cretz: Templates do not travel with the library as a dependency and templates aren't how the other files for this repo were done, they were done with strings as well (go.mod, main.go, build.gradle, settings.gradle, pyproject.toml, etc). This returns to consistency with those (unless we should turn everything to templates and then embed templates into the library for distribution, but not sure that is worth it during this refactor effort).

Sushisource: Let's just embed them then, seems easy enough. I think it's quite a bit more readable - especially this particular bit I highlighted.

cretz:
Putting in draft while I make a few fixes

sdkbuild/typescript.go
cretz: Changed logic anyways to work on Windows (properly escape backslashes in JSON strings properly) and to work with relative paths to cloned SDK directories.

cretz:
Ok, all fixed, now ready for review

sdkbuild/java.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Create build.gradle and settings.gradle
+	temporalSDKDependency := ""
+	if isPathDep {
+		temporalSDKDependency = "implementation fileTree(dir: 'sdkjars', include: ['*.jar'])"
+	} else if options.Version != "" {
+		temporalSDKDependency = fmt.Sprintf("implementation 'io.temporal:temporal-sdk:%v'",
+			strings.TrimPrefix(options.Version, "v"))
+	}
+	buildGradle := `
```
Quinn-With-Two-Ns, line 108: is `sdkbuild` is intended to be a separate library I am not sure we can hardcode all these config files

Quinn-With-Two-Ns: we allow the `go.mod` file to be customized for the builder for Go

cretz: > is sdkbuild is intended to be a separate library 

Yes

> I am not sure we can hardcode all these config files

For most, we always have, this refactor just moves it

> we allow the go.mod file to be customized for the builder for Go

Not following what you mean here

Quinn-With-Two-Ns: When this wasn't a library it didn't matter that it was hard coded, but as a library I think it does matter. You allow the `go.mod` to be customized, but not the other config files like the gradle. I am arguing all languages build config files should be customizable.

cretz: I only allow what needs to be customized to support my two needs for flexibility reasons. I would take pieces of the go.mod if that were reasonable. This library should not be considered to have a stable API. I may need to update the docs with that. This is a simple code sharing exercise between two testing tools that don't require backwards compatibility. Don't think of it as an officially distributed Go library. It won't even have a version.

I fully expect as more flexibility is needed it is added. I tried to make the most shared I could without spending a lot oftime here (which means things like package.json and pyproject.toml have a lot shared, but go.mod has project name and such that made it make less sense).

sdkbuild/typescript.go
```diff
@@ -0,0 +1,226 @@
...
+type BuildTypeScriptProgramOptions struct {
+	// Directory that will have a temporary directory created underneath.
+	BaseDir string
+	// Required version. If it contains a slash it is assumed to be a path with a
+	// package.json. Otherwise it is a specific version (with leading "v" is
+	// trimmed if present).
+	Version string
+	// Required set of paths to include in tsconfig.json paths for the harness.
+	// The paths should be relative to one-directory beneath BaseDir.
+	HarnessPaths map[string][]string
```
bergundy, line 23: Can we make this more generic and not require the concept of a harness?

sdkbuild/typescript.go
```diff
@@ -0,0 +1,226 @@
...
+			"internal-non-workflow-common", "proto", "worker", "workflow"}
+		for _, pkg := range pkgs {
+			pkgPath := "file:" + filepath.Join(localPath, "packages", pkg)
+			packageJSONDepStr += fmt.Sprintf(`"@temporalio/%v": %q,`, pkg, pkgPath)
+			packageJSONDepStr += "\n    "
+		}
+	} else {
+		packageJSONDepStr = `"temporalio": "` + strings.TrimPrefix(options.Version, "v") + "\",\n    "
+	}
+	packageJSON := `{
```
bergundy, line 112: My personal preference would be to have Go struct that's serialized as JSON but it's not like this is a regression from what we previously had, so I'm fine keeping this way.

sdkbuild/typescript.go
bergundy: Yeah, just call this `AdditionalTsConfigPaths`.

cretz: Harness is just an English term. I guess I can change it to `AdditionalTSConfigPaths`

sdkbuild/python.go
```diff
@@ -0,0 +1,154 @@
...
+type BuildPythonProgramOptions struct {
+	// Directory that will have a temporary directory created underneath. This
+	// should be a Poetry project with a pyproject.toml.
+	BaseDir string
+	// Required version. If it contains a slash it is assumed to be a path with
+	// a single wheel in the dist directory. Otherwise it is a specific version
+	// (with leading "v" is trimmed if present).
+	Version string
+	// Required Poetry dependency name of BaseDir.
+	HarnessDependency string
```
bergundy, line 24: Same as my TS comment (below), name this `AdditionalDependencies` or something similar.

bergundy (approved):
Overall LGTM

sdkbuild/typescript.go
cretz: I think it's a bit much to want to model package json as a type-safe struct in Go just as I think it's a bit much to break this little thing out into a bunch of files. YAGNI, KISS, pick your adjective. This is a couple-hundred line file (without tests) to setup a TS project so I can move on.

sdkbuild/python.go
cretz: I am trying to not be more flexible than necessary. I can change the term to `ImplementationDependency` or something, but there are no backwards compatibility requirements here forcing me to future proof this for multiple dependency names future that doesn't exist (until it does).

sdkbuild/typescript.go
cretz: The `Additional` makes it seem optional, but `TSConfigPaths`, ok

sdkbuild/typescript.go
Sushisource: Mostly my argument is I just can't see how a template is any more effort than this (it strikes me as less). I agree a struct does seem like more effort.

But, whatever, it's not terribly important and you seem to like this so works for me

Sushisource (approved):

sdkbuild/typescript.go
cretz: It's not that much more effort (it's a little more setting it up with `go:embed`, at which point I might make the template a string literal, at which point I end up with this anyways). It's just that it's arguably of not much benefit. That people have to learn the Go template language to contribute and then it was still not as featureful as the other ones (I had to do extra work to make it work with relative paths from CLI and it was not written with cross-platform in mind because the JSON wasn't escaped).

PR #270, opened 2023-05-01
Add update feature tests for Java SDK

Quinn-With-Two-Ns:
Add update feature tests for Java. I skipped interceptors because as I think I'l need to do a breaking change there.

I also updated the panic spec to reflect what we discussed on slack


features/update/updateutil/UpdateUtil.java
```diff
@@ -0,0 +1,39 @@
...
+        try {
+            var history = runner.getWorkflowHistory(run);
+            var event = history.getEventsList().stream().filter(e -> e.hasWorkflowExecutionUpdateRejectedEventAttributes()).findFirst();
+            Assertions.assertFalse(event.isPresent());
+        } catch(Exception e) {
+            Assertions.fail();
+        }
+    }
+
+    public static String CheckServerSupportsUpdate(WorkflowClient client){
```
cretz, line 23:
```suggestion
    public static String checkServerSupportsUpdate(WorkflowClient client) {
```

Is your IDE not correcting the spacing?

features/update/updateutil/UpdateUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.jupiter.api.Assertions;
+
+public class UpdateUtil {
+
+    public static void RequireNoUpdateRejectedEvents(Runner runner, Run run) {
+        try {
+            var history = runner.getWorkflowHistory(run);
+            var event = history.getEventsList().stream().filter(e -> e.hasWorkflowExecutionUpdateRejectedEventAttributes()).findFirst();
+            Assertions.assertFalse(event.isPresent());
+        } catch(Exception e) {
```
cretz, line 18: Just let it throw

features/update/activities/feature.java
```diff
@@ -0,0 +1,94 @@
...
+    public void finish() {
+      this.doFinish = true;
+    }
+
+    @Override
+    public Run execute(Runner runner) {
+      String reason = UpdateUtil.CheckServerSupportsUpdate(runner.client);
+      if (!reason.isEmpty()) {
+        runner.Skip(reason);
+      }
```
cretz, line 82: Too much Go-ism. Just add a `skipIfUpdateNotSupported` method on Runner I think.

features/update/self/feature.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @UpdateMethod()
+  void update();
+
+
+  class Impl implements feature {
+    private String state = INITIAL_STATE;
+    private WorkflowClient client;
+
+
```
cretz, line 35: Sometimes you put 2 lines between fields and methods, sometimes 1, sometimes 0

features/update/self/feature.java
```diff
@@ -0,0 +1,73 @@
...
+  @ActivityMethod
+  void selfUpdate();
+
+  @UpdateMethod()
+  void update();
+
+
+  class Impl implements feature {
+    private String state = INITIAL_STATE;
```
cretz, line 32: Sometimes you put a blank line between class decl and fields, sometimes not (I have not been applying very close scrutiny to PRs on this repo, but I am worried things are gonna get a bit bad...I wonder if we can apply a format check in CI?)

features/update/worker_restart/feature.java
```diff
@@ -0,0 +1,147 @@
...
+    private int counter = 0;
+
+    static Object updateStartedLock = new Object();
+    static Boolean updateStarted = false;
+
+    static Object updateContinueLock = new Object();
+    static Boolean updateContinue = false;
+
+    private static void signalUpdateStarted() {
+      synchronized (updateStartedLock) {
```
cretz, line 42: This isn't the best wake up approach in modern Java IMO. Consider CompletableFuture, Condition, Semaphore, etc.

features/update/worker_restart/feature.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @SignalMethod
+  void finish();
+
+  class Impl implements feature {
+
+    private boolean doFinish = false;
+    private int counter = 0;
+
+    static Object updateStartedLock = new Object();
```
cretz, line 35: Make fields final that you never mutate

features/update/worker_restart/feature.java
```diff
@@ -0,0 +1,147 @@
...
+    static Boolean updateStarted = false;
```
cretz, line 36: Do not use primitive wrappers unless you have to

features/update/worker_restart/feature.java
```diff
@@ -0,0 +1,147 @@
...
+      var stub = runner.client.newWorkflowStub(feature.class, run.execution.getWorkflowId());
+
+      CompletableFuture<Integer> updateResult = CompletableFuture.supplyAsync(() -> stub.update(1));
+      waitUpdateStarted();
+      runner.getWorkerFactory().shutdown();
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
+      }
```
cretz, line 127: Why sleep?

cretz, line 127: Just add `throws Exception` to this function and let things throw

harness/java/io/temporal/sdkfeatures/Main.java
```diff
@@ -1,29 +1,61 @@
...
+  enum Outcome {
+    PASSED,
+    FAILED,
+    SKIPPED;
+  }
+
+  class SummaryEntry {
+    String name;
+    String outcome;
+    String message;
```
cretz, line 33: Make fields private in POJOs

harness/java/io/temporal/sdkfeatures/Main.java
```diff
@@ -60,6 +92,27 @@ public void run() {
...
+      URI uri = new URI(summaryUri);
+      switch (uri.getScheme()) {
+        case "tcp":
+          Socket socket = new Socket(uri.getHost(), uri.getPort());
+          writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), "UTF-8"));
+          break;
+        case "file":
+          FileWriter fileWriter = new FileWriter(uri.getPath(), true);
+          writer = new BufferedWriter(fileWriter);
+          break;
```
cretz, line 107: You should be closing things you open. Consider a try with resources, or better, we should have used a higher level protocol like HTTP.

features/update/updateutil/UpdateUtil.java
Quinn-With-Two-Ns: all spacing looks normal in my IDE, i'll admit I didn't even consider looking at the PR for spacing. I'll look at enabling auto formating

Quinn-With-Two-Ns: Enabled spotless with the same setting as the SDK uses

features/update/worker_restart/feature.java
Quinn-With-Two-Ns: Match the Go SDK feature test

cretz: I think in Go we had the problem of worker not technically being completely stopped when Stop returned. But I don't think we have to do this here or any SDK where shutdown returning successfully guarantees worker completely stopped.

Quinn-With-Two-Ns: That is not true for Java

features/update/updateutil/UpdateUtil.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes I agree, I moved the java code. I'll move the Go in a follow up PR

features/update/worker_restart/feature.java
Quinn-With-Two-Ns: What your saying my university Java course failed me 😮, I replaced with semaphores.

features/update/worker_restart/feature.java
Quinn-With-Two-Ns: at least according to the docs for `shutdown`

Quinn-With-Two-Ns:
Ci failure looks like https://github.com/diffplug/spotless/issues/1242

features/update/activities/feature.java
```diff
@@ -0,0 +1,88 @@
...
+    public void workflow() {
+      Workflow.await(() -> this.doFinish);
+    }
+
+    @Override
+    public int update() {
+      var activities =
+          activities(
+              feature.class, builder -> builder.setScheduleToCloseTimeout(Duration.ofSeconds(5)));
+      List<String> results = new ArrayList();
```
cretz, line 48:
```suggestion
      List<String> results = new ArrayList<>();
```

harness/java/io/temporal/sdkfeatures/Main.java
```diff
@@ -1,29 +1,79 @@
...
+    public String getOutcome() {
+      return outcome;
+    }
+
+    public String getMessage() {
+      return message;
+    }
+  }
+
+  BufferedWriter createSummaryServerWriter() {
```
cretz, line 54: Ug, I still wish this was an HTTP server (metadata, status codes, content type, well supported, concurrent, req/resp if we want it, etc). But maybe we can tackle that at a later time.

cretz (approved):
LGTM, only cosmetic stuff

PR #272, opened 2023-05-17
An sdk-go feature test for async (accepted) update

mmcshane:
## What was changed

An sdk-go feature test for async (accepted) update

## Why?

Server supports async-accepted now (with appropriate dynamicconfig) so let's test it.

## Checklist

1. Closes 

2. How was this tested:

Ran with dynamicconfig `frontend.enableUpdateWorkflowExecutionAsyncAccepted=true`

3. Any docs updates needed?

No

### DIFF
--- a/features/update/async_accepted/README.md
+++ b/features/update/async_accepted/README.md
@@ -1,0 +1,18 @@
+# Async Updates
+
+Updates can be invoked asynchronously by waiting on the ACCEPTED state.
+
+# Detailed spec
+
+It will be possible to invoke an update and indicate that the calling RPC client
+wants the RPC call to block on on completion of the update but rather on the the
+update passing validation on the workflow - this is called the "Accepted" state.
+The update continues to execute to completion despite there being no caller
+waiting on its outcome.
+
+A client with knowledge of the update's identifying information (viz.,
+workflow ID, run ID, and update ID) can create a handle to that update and await
+the outcome of the update call. Outcomes returned from these asynchronous
+updates are the same as those that would have been returned inline had the
+update been invoked synchronously.
+
--- a/features/update/async_accepted/feature.go
+++ b/features/update/async_accepted/feature.go
@@ -1,0 +1,165 @@
+package async_accepted
+
+import (
+	"context"
+	"errors"
+	"time"
+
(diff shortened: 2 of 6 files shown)


features/update/async_accepted/feature.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, runner.Skip(reason)
+		}
+		run, err := runner.ExecuteDefault(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		updateCounter := 0
+		requestedSleep := 2 * time.Second
+		sendAnAsyncUpdate := func(
```
cretz, line 44: I usually don't care too much about code organization during tests, but I think people are gonna read these features as how-to's. I think we should just inline this call each time needed. I know it kinda violates reuse, but this is a two-statement helper and arguably the counter is not needed.

features/update/async_accepted/feature.go
```diff
@@ -0,0 +1,134 @@
...
+					Args:       []interface{}{requestedSleep, disp},
+					WaitPolicy: &updatepb.WaitPolicy{
+						LifecycleStage: enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED,
+					},
+				},
+			)
+		}
+
+		start := time.Now()
+		originalHandle, err := sendAnAsyncUpdate(ctx, succeed)
```
cretz, line 65: If possible, again because I and others are kinda reading this as a how-to currently, can you add a simple comment above here and before `sendAnAsyncUpdate` in sections below on the situation you're testing. E.g.

```suggestion
		// Simple accepted update should return quickly not wait for completion
		originalHandle, err := sendAnAsyncUpdate(ctx, succeed)
```

cretz (approved):
Looks good, couple of suggestions but no blockers

features/update/async_accepted/README.md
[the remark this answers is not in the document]
mmcshane: For the purposes of the features repo that's definitely a separate feature. It will return Unimplemented from the server right now so I guess technically I could write a test that will always get skipped? Other priorities for now.

PR #275, opened 2023-05-25
Modify update tests to require correct panic behavior

mmcshane:
## What was changed

Aligned tests to expected behaviors and updated go.mod to actually get that behavior in the sdk.

## Why?

Update validation panics should fail the update. Update execution panics should abide by the WorkflowPanicPolicy.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


features/features.go
```diff
@@ -60,13 +60,13 @@ func init() {
...
-		update_async_accepted.Feature,
 		update_activities.Feature,
-		update_task_failure.Feature,
+		update_async_accepted.Feature,
 		update_intercept.Feature,
-		update_validation_replay.Feature,
+		update_non_durable_reject.Feature,
 		update_self.Feature,
+		update_task_failure.Feature,
+		update_validation_replay.Feature,
```
mmcshane, line 69: I sorted these alphabetically, sorry for the noise

features/update/async_accepted/feature.go
```diff
@@ -24,7 +24,6 @@ const (
 
 	succeed       UpdateDisposition = 0
 	failWithError UpdateDisposition = 1
-	failWithPanic UpdateDisposition = 2
```
mmcshane, line 27: This now hangs or fails the workflow. Either way it doesn't fail the update anymore so removed form this test.

features/update/task_failure/feature.go
```diff
@@ -2,21 +2,29 @@ package task_failure
...
+	panicFromExecUpdate     = "panic_from_exec"
+	panicFromValidateUpdate = "panic_from_validate"
+	shutdownSignal          = "shutdown!"
 )
 
 var Feature = harness.Feature{
-	Workflows: Workflow,
+	Workflows:                          PanickyUpdates,
+	WorkerOptions:                      worker.Options{WorkflowPanicPolicy: worker.BlockWorkflow},
+	DisableWorkflowPanicPolicyOverride: true,
```
cretz, line 27: :+1: I think (hope) this will be the only feature setting this anyways and I think is a great use of feature options (I expect the feature options list to grow and grow for these kinds of needs).

cretz (approved):

PR #282, opened 2023-06-09
Add data_converter features in go (#81)

antlai-temporal:
Add feature tests for protobuf, json_protobuf, json, compression, and encryption.

Includes only implementations in go, other languages will follow

## What was changed

Add go feature tests for data_converter, first part of #81
 
## Checklist

1. Closes 
Progress towards #81


features/data_converter/encryption/data_converter.go
cretz, line 1: Arguably this feature is not testing a feature of our SDK anymore than compression is. It's more of a sample.

features/data_converter/json_protobuf/feature.go
```diff
@@ -0,0 +1,64 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+var EXPECTED_RESULT = []byte{0xde, 0xad, 0xbe, 0xef}
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	CheckResult: CheckResult,
+	ClientOptions: client.Options{
+		DataConverter: converter.NewCompositeDataConverter(
```
cretz, line 21: You don't need a custom data converter, this feature is on by default in our SDKs

harness/go/harness/util.go
```diff
@@ -41,6 +41,29 @@ func FindEvent(history client.HistoryEventIterator, cond func(*historypb.History
 	return nil, nil
 }
 
+// GetWorkflowResultPayload returns the first payload of a workflow result
+func GetWorkflowResultPayload(ctx context.Context, client client.Client,
+	workflowId string) (*common.Payload, error) {
```
cretz, line 46:
```suggestion
func GetWorkflowResultPayload(
	ctx context.Context,
	client client.Client,
	workflowID string,
) (*common.Payload, error) {
```

If you must wrap, this is usually the better way. And "ID" in Go is usually capitalized (when not part of generated code).

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"github.com/gogo/protobuf/proto"
+	pb "go.temporal.io/features/features/data_converter"
+	"go.temporal.io/features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/workflow"
+)
+
+var EXPECTED_RESULT = []byte{0xde, 0xad, 0xbe, 0xef}
```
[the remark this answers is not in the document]
antlai-temporal: Changed to camelCase

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
+package binary_protobuf
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/gogo/protobuf/proto"
```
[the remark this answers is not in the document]
antlai-temporal: Switched to goimports and that fixed it, thanks!

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	pb "go.temporal.io/features/features/data_converter"
```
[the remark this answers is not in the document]
antlai-temporal: Good point. I found DataBlob to be almost identical to the one I created, and reused it.

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	CheckResult: CheckResult,
+	ClientOptions: client.Options{
+		DataConverter: converter.NewCompositeDataConverter(
+			converter.NewNilPayloadConverter(),
+			// Disable ByteSlice and ProtoJSON converters
+			converter.NewProtoPayloadConverter(),
+			converter.NewJSONPayloadConverter(),
```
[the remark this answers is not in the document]
antlai-temporal: Done

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	if !bytes.Equal(result.Data, EXPECTED_RESULT) {
+		return fmt.Errorf("invalid result: %v", result)
+	}
+
+	payload, err := harness.GetWorkflowResultPayload(ctx, runner.Client, run.GetID())
+	if err != nil {
+		return err
+	}
+
+	var encoding = string(payload.GetMetadata()["encoding"])
```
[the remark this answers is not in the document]
antlai-temporal: Yes, good idea. Done.

features/data_converter/encryption/data_converter.go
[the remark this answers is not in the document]
antlai-temporal: I added a subdir for the custom encryption codec.

antlai-temporal: I think what is showing is that we can add custom converters, all the others are already part of the SDK...

features/data_converter/json_protobuf/feature.go
antlai-temporal: I added a comment to clarify that ProtoJSONPayload has priority over ProtoPayload in the default converter, and deleted it.

harness/go/harness/util.go
antlai-temporal: Much cleaner. Is there are guideline for max columns?, I'm assuming about 120...

harness/go/harness/util.go
```diff
@@ -41,6 +41,29 @@ func FindEvent(history client.HistoryEventIterator, cond func(*historypb.History
...
+	history := client.GetWorkflowHistory(ctx, workflowId, "", false, 0)
```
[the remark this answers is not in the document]
antlai-temporal: Done

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	resultInHistory := pb.BinaryMessage{}
+	if err := proto.Unmarshal(payload.GetData(), &resultInHistory); err != nil {
+		return err
+	}
+
+	if !bytes.Equal(resultInHistory.GetData(), EXPECTED_RESULT) {
+		return fmt.Errorf("invalid result in history: %v", resultInHistory)
+	}
+
+	runner.Require.Equal(result, resultInHistory)
```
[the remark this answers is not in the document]
antlai-temporal: Makes sense, thanks. Done.

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+		DataConverter: converter.NewCompositeDataConverter(
+			converter.NewNilPayloadConverter(),
+			// Disable ByteSlice and ProtoJSON converters
+			converter.NewProtoPayloadConverter(),
+			converter.NewJSONPayloadConverter(),
+		),
+	},
+}
+
+func Workflow(ctx workflow.Context) (pb.BinaryMessage, error) {
```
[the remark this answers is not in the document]
antlai-temporal: The default `runner.Execute` does not take arguments for a workflow. We could do it with a custom `Execute`, or a child workflow. I think both would complicate the feature though, and with child workflows you always have the doubt of whether it only applies to child workflows... 

I'm leaning  towards opening an issue to add default workflow arguments in ExecuteDefault, and then get back to this when it is supported...

features/data_converter/compression/feature.go
```diff
@@ -0,0 +1,88 @@
...
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	CheckResult: CheckResult,
+	ClientOptions: client.Options{
+		DataConverter: getCompressionConverter(),
+	},
+}
+
+func Workflow(ctx workflow.Context) (Message, error) {
```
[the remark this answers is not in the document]
antlai-temporal: See previous comment

features/data_converter/binary_protobuf/feature.go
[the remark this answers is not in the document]
cretz: I think you just happen to be the first person to need arguments for a workflow. No need to wait for another issue. Can do this generically, but I think something like this is simple enough (unchecked just typed here, so may have typos):

```
Execute: func(ctx context.Context, r *Runner) (client.WorkflowRun, error) {
    opts := r.Feature.StartWorkflowOptions
    opts.TaskQueue = r.TaskQueue
    return r.Client.ExecuteWorkflow(ctx, opts, Workflow, &MyWorkflowParam{})
},
```

That's only a few lines.

features/data_converter/encryption/data_converter.go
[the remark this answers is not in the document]
cretz: I still think it's better if it's in this file personally, but meh. Hopefully you can keep it in a single file when you get to the other languages.

cretz: Not following. This shows how to use `converter.NewCodecDataConverter` just like `compression`. So what is this sample showing that the `compression` sample doesn't?

I think our feature should be "codec" and not "compression" and "encryption" which are just specializations of that. I don't think we even need a feature to require every SDK have a Zlib codec even though Go and Java might happen to. We just need a feature showing how to use a custom codec (whatever that custom codec does).

Unfortunately I know these directories were already here, but I think we should remove `compression` and `encryption` directories and just have a `codec` directory. Can wait for others' opinions though.

harness/go/harness/util.go
cretz: No strict guideline in Go, but yeah 120 is reasonable (assuming tab width at 2, different people have different widths)

features/data_converter/binary_protobuf/README.md
```diff
@@ -2,15 +2,17 @@
...
+- verify payload encoding is `binary/protobuf`, unmarshall its data into a
+`DataBlob` using `protobuf` library, and compare it to the client
+result
 
 # Detailed spec
 
 `metadata.encoding = toBinary("binary/protobuf")`
-`metadata.messageType = toBinary("BinaryMessage")` (used by languages that cannot get a parameter's type at runtime)
\ No newline at end of file
+`metadata.messageType = toBinary("DataBlob")` (used by languages that cannot get a parameter's type at runtime)
```
bergundy, line 18:
```suggestion
`metadata.messageType = toBinary("temporal.api.common.v1.DataBlob")` (used by languages that cannot get a parameter's type at runtime)
```

features/data_converter/encryption/codecs/data_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	"io"
+
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/converter"
+)
+
+const (
+	MetadataEncodingEncrypted = "binary/encrypted"
+	MetadataEncryptionKeyID   = "encryption-key-id"
+	TEST_KEY_STR              = "test-key-test-key-test-key-test!"
```
bergundy, line 19:
```suggestion
	TestKeyStr                = "test-key-test-key-test-key-test!"
```

features/data_converter/encryption/codecs/data_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return &DataConverter{
+		DataConverter: converter.NewCodecDataConverter(dataConverter, codecs...),
+		options:       options,
+	}
+}
+
+// Codec implements PayloadCodec using AES Crypt.
+type Codec struct {
```
bergundy, line 44: Does this need to be public?

features/data_converter/encryption/codecs/data_converter.go
```diff
@@ -0,0 +1,150 @@
...
+	KeyID string
+}
+
+func (e *Codec) getKey(keyID string) (key []byte) {
```
bergundy, line 48: Not sure why you need this method.
In a real payload the signature for getting a key would return an error.
Also note the unused param here, maybe return []byte(keyID) (plus some prefix / suffix).

features/data_converter/json_protobuf/README.md
```diff
@@ -2,14 +2,16 @@
...
+- verify payload encoding is `json/protobuf`, unmarshall its data into a
+`DataBlob` using `jsonpb` library, and compare it to the client
+result
 
 # Detailed spec
 
 `metadata.encoding = toBinary("json/protobuf")`
-`metadata.messageType = toBinary("BinaryMessage")` (used by languages that cannot get a parameter's type at runtime)
\ No newline at end of file
+`metadata.messageType = toBinary("DataBlob")` (used by languages that cannot get a parameter's type at runtime)
```
bergundy, line 17: Should be the fully qualified message type.

features/data_converter/json_protobuf/feature.go
```diff
@@ -0,0 +1,55 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+var expectedResult = commonpb.DataBlob{Data: []byte{0xde, 0xad, 0xbe, 0xef}}
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	CheckResult: CheckResult,
+	// No need of a custom data converter, the default one prioritizes
+	//  ProtoJSONPayload over ProtoPayload
```
bergundy, line 21: nit:

```suggestion
	// ProtoJSONPayload over ProtoPayload
```

bergundy (approved):
Only added some very minor comments.
This is good to merge AFAIC once those are addressed.

features/data_converter/binary_protobuf/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	return pb.BinaryMessage{Data: EXPECTED_RESULT}, nil
+}
+
+func CheckResult(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) error {
+	// verify client result is BinaryMessage `0xdeadbeef`
+	result := pb.BinaryMessage{}
+	if err := run.Get(ctx, &result); err != nil {
+		return err
+	}
+	if !bytes.Equal(result.Data, EXPECTED_RESULT) {
```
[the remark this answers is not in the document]
antlai-temporal: Done

features/data_converter/encryption/data_converter.go
antlai-temporal: I see what you mean. It will certainly simplify things if we call it codec and just use the built in compression, the custom encryption codecs are just examples that add clutter. @bergundy what do you think, shall we get rid of encryption?

cretz: Not even built-in, you can just make a simple base64 codec that serializes and base64's the payload proto. I do this in .NET for example: https://github.com/temporalio/sdk-dotnet/blob/main/tests/Temporalio.Tests/Converters/Base64PayloadCodec.cs

antlai-temporal: Makes sense. If @bergundy is happy with this, I will rename it to codec, remove encryption, and if there is a built in codec in the SDK, e.g., compression in go, I'll just use it, otherwise just create a simple base64 codec in the same file

bergundy: TBH I was surprised that we have separated this into specific features when I initially read it but found it nice to have some (somewhat) samples in features.

Since we already cover compression in our samples repos (where our users should be looking), I'd be fine with showing codec as a concept instead of encryption / compression separately.

cretz: :+1: Yeah, I think we should let features be more about the features and less about what you can do with the features.

features/data_converter/binary_protobuf/feature.go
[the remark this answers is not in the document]
antlai-temporal: Adding an echo workflow to test in both directions. 
Replay was failing with a custom codec because the DataConverter option was not propagated to WorkflowReplay, fixed in runner.go, please check.
It was a great idea to add this extra check, this was not showing up before...

features/data_converter/encryption/data_converter.go
antlai-temporal: Replaced encryption and compression with a generic codec (internally base64 encoding as suggested...)

features/data_converter/binary_protobuf/README.md
antlai-temporal: Done

features/data_converter/json_protobuf/README.md
antlai-temporal: Done

features/data_converter/json_protobuf/feature.go
antlai-temporal: Fixed

harness/go/harness/util.go
```diff
@@ -41,6 +42,72 @@ func FindEvent(history client.HistoryEventIterator, cond func(*historypb.History
...
+		return nil, err
+	}
+	if event == nil {
+		return nil, errors.New("missing WorkflowExecutionCompleted event")
+	}
+
+	attrs := event.GetWorkflowExecutionCompletedEventAttributes()
+	if attrs == nil {
+		return nil, errors.New("could not locate WorkflowExecutionCompleted event")
+	}
```
cretz, line 81: Not really possible, but meh

harness/go/harness/util.go
```diff
@@ -41,6 +42,72 @@ func FindEvent(history client.HistoryEventIterator, cond func(*historypb.History
...
+	return attrs.GetResult().GetPayloads()[0], nil
```
cretz, line 82: You can use the fields when you are not concerned about nil safety (e.g. `attrs.Result.Payloads[0]`), but you may want to check for nil result or empty payloads. No big deal though.

features/data_converter/codec/feature.go
```diff
@@ -0,0 +1,142 @@
...
+		b, err := proto.Marshal(p)
+		if err != nil {
+			return payloads, err
+		}
+		dst := make([]byte, base64.StdEncoding.EncodedLen(len(b)))
+		base64.StdEncoding.Encode(dst, b)
+		result[i] = &commonpb.Payload{
+			Metadata: map[string][]byte{converter.MetadataEncoding: []byte("my-encoding")},
+			Data:     dst,
+		}
```
cretz, line 116:
```suggestion
		result[i] = &commonpb.Payload{
			Metadata: map[string][]byte{converter.MetadataEncoding: []byte("my-encoding")},
			Data:     []byte(base64.StdEncoding.EncodeToString()),
		}
```

Just to make it easier on you

features/data_converter/codec/feature.go
```diff
@@ -0,0 +1,142 @@
...
+// A simple codec that encodes binary arrays into Base64.
+// The encoding type is "my-encoding", representing an arbitrary custom codec.
+func newBase64Codec() converter.PayloadCodec { return &base64Codec{} }
+
+func (b *base64Codec) Encode(payloads []*commonpb.Payload) ([]*commonpb.Payload, error) {
+	result := make([]*commonpb.Payload, len(payloads))
+	for i, p := range payloads {
+		b, err := proto.Marshal(p)
+		if err != nil {
+			return payloads, err
```
cretz, line 109:
```suggestion
			return nil, err
```

features/data_converter/codec/feature.go
```diff
@@ -0,0 +1,142 @@
...
+		if string(p.Metadata[converter.MetadataEncoding]) != "my-encoding" {
+			result[i] = p
+			continue
+		}
+		dst := make([]byte, base64.StdEncoding.DecodedLen(len(p.Data)))
+		n, err := base64.StdEncoding.Decode(dst, p.Data)
+		if err != nil {
+			return payloads, err
+		}
+		dst = dst[:n]
```
cretz, line 134:
```suggestion
		dst, err := base64.StdEncoding.DecodeString(string(p.Data))
		if err != nil {
			return payloads, err
		}
```

harness/go/harness/runner.go
```diff
@@ -247,7 +247,19 @@ func (r *Runner) CheckHistoryDefault(ctx context.Context, _ client.WorkflowRun)
...
+		var err error
+		replayer, err = worker.NewWorkflowReplayerWithOptions(
+			worker.WorkflowReplayerOptions{DataConverter: dataConverter},
+		)
+		if err != nil {
+			return err
+		}
+	} else {
+		replayer = worker.NewWorkflowReplayer()
+	}
```
cretz, line 262:
```suggestion
	replayer, err = worker.NewWorkflowReplayerWithOptions(
		worker.WorkflowReplayerOptions{DataConverter: r.Feature.ClientOptions.DataConverter},
	)
	if err != nil {
		return err
	}
```

Nil option means default

cretz (approved):
Looks good, nothing blocking

features/data_converter/codec/feature.go
antlai-temporal: Yes, EncodeToString is much clear

features/data_converter/codec/feature.go
antlai-temporal: Yes, DecodeString is easier...

harness/go/harness/runner.go
antlai-temporal: Fixed

Issue #290, opened 2023-06-28
[Feature Request] Add feature that tests for update deduplication


antlai-temporal:
Updates that reuse the update ID, and target the same workflow, should only be performed once.

Add Java and Go features that show update deduplication.

Linked to JIRA OSS-1250

PR #291, opened 2023-06-28
Add features that test for update deduplication (#290)

antlai-temporal:
Test that updates with the same update ID, and targeting the same workflow instance,  are properly deduplicated.

Closes Issue #290 and Jira OSS-1250

### DIFF
--- a/features/update/deduplication/feature.go
+++ b/features/update/deduplication/feature.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"time"
 
+	enumspb "go.temporal.io/api/enums/v1"
+	updatepb "go.temporal.io/api/update/v1"
 	"go.temporal.io/features/features/update/updateutil"
 	"go.temporal.io/features/harness/go/harness"
 	"go.temporal.io/sdk/client"
@@ -35,6 +37,9 @@ var Feature = harness.Feature{
 				WorkflowID: run.GetID(),
 				RunID:      run.GetRunID(),
 				UpdateName: incrementCount,
+				WaitPolicy: &updatepb.WaitPolicy{
+					LifecycleStage: enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED,
+				},
 			},
 		)
 		runner.Require.NoError(err)
@@ -46,6 +51,9 @@ var Feature = harness.Feature{
 				WorkflowID: run.GetID(),
 				RunID:      run.GetRunID(),
 				UpdateName: incrementCount,
+				WaitPolicy: &updatepb.WaitPolicy{
+					LifecycleStage: enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED,
+				},
 			},
 		)
 		runner.Require.NoError(err)
--- a/features/update/deduplication/feature.go
+++ b/features/update/deduplication/feature.go
@@ -60,6 +60,12 @@ var Feature = harness.Feature{
 		runner.Require.NoError(runner.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), shutdownSignal, nil))
 		updateutil.RequireNoUpdateRejectedEvents(ctx, runner)
 
+		nUpdates, err := harness.GetCountCompletedUpdates(ctx, runner.Client, run.GetID())
+		if err != nil {
+			return nil, err
+		}
+		runner.Require.Equal(expectedCount, nUpdates)
+
 		return run, ctx.Err()
 	},
 }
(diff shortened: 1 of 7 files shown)


features/update/deduplication/feature.go
```diff
@@ -0,0 +1,90 @@
...
+
+		var result int
+		runner.Require.NoError(handle1.Get(ctx, &result))
+		runner.Require.Equal(expectedCount, result)
+
+		runner.Require.NoError(handle2.Get(ctx, &result))
+		runner.Require.Equal(expectedCount, result)
+
+		runner.Require.NoError(runner.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), shutdownSignal, nil))
+		updateutil.RequireNoUpdateRejectedEvents(ctx, runner)
```
cretz, line 61: Do you need to confirm that only one update appeared in history?

features/update/deduplication/feature.go
```diff
@@ -0,0 +1,90 @@
...
+		handle2, err := runner.Client.UpdateWorkflowWithOptions(
+			ctx,
+			&client.UpdateWorkflowWithOptionsRequest{
+				UpdateID:   reusedUpdateID,
+				WorkflowID: run.GetID(),
+				RunID:      run.GetRunID(),
+				UpdateName: incrementCount,
+			},
+		)
+		runner.Require.NoError(err)
```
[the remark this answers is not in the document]
antlai-temporal: I think you are right about idempotence.  I can see an activity doing an update, and if something else after that goes wrong, all the later retries of the activity would fail if they don't ignore the error...

features/update/deduplication/feature.go
```diff
@@ -0,0 +1,90 @@
+package deduplication
+
+import (
+	"context"
+	"time"
+
+	"go.temporal.io/features/features/update/updateutil"
```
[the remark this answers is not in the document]
antlai-temporal: Happy to do another PR later on to restructure this one and #252, if @mmcshane is ok with that...

features/update/deduplication/feature.go
antlai-temporal: Good idea, done.

features/update/deduplication/feature.go
```diff
@@ -0,0 +1,96 @@
...
+
+		var result int
+		runner.Require.NoError(handle1.Get(ctx, &result))
+		runner.Require.Equal(expectedCount, result)
+
+		runner.Require.NoError(handle2.Get(ctx, &result))
+		runner.Require.Equal(expectedCount, result)
+
+		runner.Require.NoError(runner.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), shutdownSignal, nil))
+		updateutil.RequireNoUpdateRejectedEvents(ctx, runner)
```
Quinn-With-Two-Ns, line 61: How come the Java test does not check any history?

features/update/deduplication/feature.go
```diff
@@ -0,0 +1,96 @@
...
+	Execute: func(ctx context.Context, runner *harness.Runner) (client.WorkflowRun, error) {
+		if reason := updateutil.CheckServerSupportsUpdate(ctx, runner.Client); reason != "" {
+			return nil, runner.Skip(reason)
+		}
+		run, err := runner.ExecuteDefault(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		handle1, err := runner.Client.UpdateWorkflowWithOptions(
```
Quinn-With-Two-Ns, line 31: Since we are using the handle API is the intention for the updates to run in parallel? Because right now I don't think they will

antlai-temporal: By adding the Sleep() in the middle of the update and using handlers I was trying to show that deduplication works even if the previous update has not completed yet. Unfortunately, you are right, there is no overlap, even if I use different IDs and different handler functions the server serializes all the updates for a workflow, i.e., one starts after the previous one completes. 

I couldn't find in the docs that this is expected behavior, but it has big implications for the programmers abstraction.  Reentrant update calls will deadlock, i.e., calling an update from an activity triggered by an update, but there are no races, which is really nice... 

In any case the docs should be explicit about this...

Quinn-With-Two-Ns: I think the reason your seeing update requests is because in `UpdateWorkflowWithOptions` you didn't set a wait policy so I think it will default to `UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED` so your waiting for the whole request to complete when you send the update, not when your waiting on the handle result

antlai-temporal: Thanks, once I set it to ACCEPTED they start to overlap (with different ids). 
What is very confusing is that in go the default is COMPLETED and  in java the default is ACCEPTED. We need to agree on one default...

PR #292, opened 2023-07-04
Add data_converter features in Java (#81)

antlai-temporal:
## What was changed
Adding Java data converter features.

1. Closes 

Progress to closing #81


features/data_converter/binary/feature.java
```diff
@@ -1,34 +1,30 @@
...
 import java.nio.file.Files;
 import java.nio.file.Paths;
 
-import com.google.protobuf.util.JsonFormat;
-
 @WorkflowInterface
 public interface feature extends Feature {
-    static byte[] deadbeef = new byte[]{(byte)0xde, (byte)0xad, (byte)0xbe, (byte)0xef};
-    
+  static byte[] deadbeef = new byte[] {(byte) 0xde, (byte) 0xad, (byte) 0xbe, (byte) 0xef};
```
Quinn-With-Two-Ns, line 18: nit: `static` is redundant here, also this variable is a constant so convention is all caps with underscore to separate words

features/data_converter/binary/feature.java
```diff
@@ -1,34 +1,30 @@
...
+
   @WorkflowMethod
   public byte[] workflow();
```
Quinn-With-Two-Ns, line 21: bit: `public` is redundant here

features/data_converter/binary/feature.java
Quinn-With-Two-Ns: I know you didn't write this code, but you did the same thing in some of the other features. Not a blocker

Quinn-With-Two-Ns (approved):

features/data_converter/binary/feature.java
antlai-temporal: Thanks, it makes sense, done...

features/data_converter/binary/feature.java
antlai-temporal: Yes, it is an interface... Done

PR #311, opened 2023-07-14
Add feature: signal child workflow

dandavison:
Implements `features/child_workflow/signal` for the 4 languages:

- A child workflow is defined that waits for a signal and then returns the signal data
- A parent workflow is defined that starts the child workflow, sends the signal, and returns the result of the child workflow.
- We assert that the result of the parent workflow is equal to the signal data that it sent.

I've checked that the workflows do not complete if they are altered to not send the signal.

I'll look into refactoring to add shared utility functions in a separate PR.

### DIFF
--- a/features/child_workflow/signal/feature.go
+++ b/features/child_workflow/signal/feature.go
@@ -1,0 +1,52 @@
+package signal
+
+import (
+	"context"
+	"time"
+
+	"go.temporal.io/features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	UnblockSignal  = "unblock-signal"
+	UnblockMessage = "unblock"
+)
+
+// A workflow that starts a child workflow, unblocks it, and returns the result
+// of the child workflow.
+func Workflow(ctx workflow.Context) (string, error) {
+	ctx = workflow.WithChildOptions(ctx, workflow.ChildWorkflowOptions{
+		WorkflowExecutionTimeout: 10 * time.Minute,
+		WorkflowTaskTimeout:      time.Minute,
+	})
+	childWorkflowFut := workflow.ExecuteChildWorkflow(ctx, ChildWorkflow)
+	childWorkflowFut.SignalChildWorkflow(ctx, UnblockSignal, UnblockMessage)
+	result := ""
+	err := childWorkflowFut.Get(ctx, &result)
+	if err != nil {
+		return "", err
+	}
+	return result, nil
+}
+
+// A workflow that waits for a signal and returns the data received.
+func ChildWorkflow(ctx workflow.Context) (string, error) {
+	unblockMessage := ""
+	signalCh := workflow.GetSignalChannel(ctx, UnblockSignal)
+	signalCh.Receive(ctx, &unblockMessage)
+	return unblockMessage, nil
+}
+
+var Feature = harness.Feature{
+	Workflows:       []interface{}{Workflow, ChildWorkflow},
+	ExpectRunResult: UnblockMessage,
+	Execute: func(ctx context.Context, runner *harness.Runner) (client.WorkflowRun, error) {
+		run, err := runner.ExecuteDefault(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return run, nil
+	},
+}
--- a/features/child_workflow/signal/feature.java
+++ b/features/child_workflow/signal/feature.java
@@ -1,0 +1,98 @@
+package child_workflow.signal;
+
+import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowOptions;
+import io.temporal.sdkfeatures.Assertions;
+import io.temporal.sdkfeatures.Feature;
+import io.temporal.sdkfeatures.Run;
+import io.temporal.sdkfeatures.Runner;
+import io.temporal.worker.Worker;
+import io.temporal.workflow.Async;
+import io.temporal.workflow.Promise;
+import io.temporal.workflow.SignalMethod;
+import io.temporal.workflow.Workflow;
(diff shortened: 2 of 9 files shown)


features/child_workflow/signal/ChildWorkflow.java
cretz, line 1: To stick with the single-file approach, can we make this an inner interface in the Java feature class? If that means making the child workflow a separate inner class too because you can't implement an outer class from an inner interface, so be it.

features/child_workflow/signal/feature.go
```diff
@@ -0,0 +1,52 @@
...
+		WorkflowTaskTimeout:      time.Minute,
+	})
+	childWorkflowFut := workflow.ExecuteChildWorkflow(ctx, ChildWorkflow)
+	childWorkflowFut.SignalChildWorkflow(ctx, UnblockSignal, UnblockMessage)
+	result := ""
+	err := childWorkflowFut.Get(ctx, &result)
+	if err != nil {
+		return "", err
+	}
+	return result, nil
```
cretz, line 31:
```suggestion
	var result string
	err := childWorkflowFut.Get(ctx, &result)
	return result, err
```

Subjective, no need to change anything

features/child_workflow/signal/feature.java
```diff
@@ -0,0 +1,80 @@
...
+      return childResult.get();
+    }
+
+    /*
+     * Child workflow
+     *
+     * A workflow that waits for a signal and returns the data received.
+     */
+
+    private String childWorkflowUnblockMessage = null;
```
cretz, line 46:
```suggestion
    private String childWorkflowUnblockMessage;
```

(nit, can ignore)

cretz (commented):
This is great! Only request before approval is to move that Java child workflow file inside the features file if we can. Awesome work. I really appreciate the simplicity and straightforwardness.

features/child_workflow/signal/ChildWorkflow.java
dandavison: OK great I'm glad you picked that up because I was also uncomfortable with the way the Java child workflow implementation was intermingled with the parent workflow impl and had to use contrived method names. I was following [child_workflow/result/feature.java](https://github.com/temporalio/features/blob/main/features/child_workflow/result/feature.java) and [child_workflow/result/ChildWorkflow.java](https://github.com/temporalio/features/blob/main/features/child_workflow/result/ChildWorkflow.java) so if I can do it along the lines you suggest maybe I'll do it for both.

features/child_workflow/signal/feature.go
dandavison: I like the visually cleaner style of that change. I don't have much Go experience and have a question: so, if there's an error, then it's possible that the data written to `&result` is corrupt or partially written, etc. Is it considered good Go philosophy to say "There was an error. I'm giving you your requested data as well, but all bets are off regarding it -- e.g. it may not even be a valid instance of its type".

dandavison: E.g. languages like Rust with result ADTs don't do that, but in languages with exceptions the user can provide a string representation of the result.

features/child_workflow/signal/ChildWorkflow.java
cretz: I disagree w/ the other one too but I didn't review it, heh :-) Single file is ideal IMO, feel free to do to both if you want, though not required.

features/child_workflow/signal/feature.go
cretz: It's a good question. Often it is implied in Go that "if there are two return types and the second is an error, they are mutually exclusive and err != nil means result has no value unless documented otherwise". But that's not a hardfast rule in all of Go. But I can say result and error are mutually exclusive for Temporal workflows, so a non-nil error discards results.

features/child_workflow/signal/ChildWorkflow.java
```diff
@@ -0,0 +1,15 @@
+package child_workflow.signal;
+
+import io.temporal.workflow.SignalMethod;
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
+
+@WorkflowInterface
+public interface ChildWorkflow {
```
cretz, line 8: Let's throw this interface as an inner interface in that `feature` class to if we can

features/child_workflow/signal/ChildWorkflow.java
```diff
@@ -0,0 +1,15 @@
...
+
+  @WorkflowMethod
+  String childWorkflowExecute();
+
+  @SignalMethod
+  public void childWorkflowUnblock(String message);
```
cretz, line 14:
```suggestion
  void childWorkflowUnblock(String message);
``

(subjective, but public is implied here)

dandavison: done

features/child_workflow/signal/ChildWorkflow.java
dandavison: Done.

features/child_workflow/signal/feature.go
dandavison: OK thanks for discussing! I think I'll leave this as-is for now and keep the issue in mind as I get more Go experience. So I'm going to provisionally say that I'm going to avoid doing it in case it gives a reader the impression that corrupt data might be returned.

features/child_workflow/signal/ChildWorkflow.java
dandavison: OK single-file version is pushed. Thanks @cretz for offline discussion leading to the new `prepareWorker` configuration point added.

cretz (approved):
Looks great! (merge at will)

dandavison:
Thanks for the review and discussions @cretz.

PR #313, opened 2023-07-15
Refactoring: shared utilities and cleanup

dandavison:
This is pure refactoring aimed at slightly reducing friction when adding new features:

- New harness utilities for starting a workflow with default options: `start_parameterless_workflow` and `executeParameterlessWorkflow` for Python and Typescript respectively.

- Convert `child_workflow/result` Java feature to use single `feature.java` file (and move `ChildWorkflow` impls inside `ChildWorkflow` interface)

- Make use of `runner.executeSingleParameterlessWorkflow` in a Java feature that wasn't using it.


features/child_workflow/signal/feature.java
```diff
@@ -13,37 +11,36 @@
...
+        return childWorkflowUnblockMessage;
+      }
 
-    @Override
-    public void unblock(String message) {
-      childWorkflowUnblockMessage = message;
+      @Override
+      public void unblock(String message) {
+        childWorkflowUnblockMessage = message;
+      }
```
dandavison, line 43: The diff here is mostly indentation, moving an impl into the interface - easier to understand with hide-whitespace on.

features/child_workflow/result/feature.java
```diff
@@ -5,52 +5,66 @@
 import io.temporal.sdkfeatures.Feature;
```
dandavison, line 5: The diff in the file is easier to understand with hide-whitespace on: spotless converting 4-space to 2-space indentation.

dandavison:
I briefly looked into changing the Java harness so that we can define multiple workflows symmetrically (rather than having one standalone workflow interface and the other workflow interface implemented by the feature), but it's fairly deeply engrained in the harness design so I left that.

features/child_workflow/result/feature.java
cretz: Greatly appreciated! I think on the Java SDK there is an auto-formatter. I wonder if we want to bring that over here.

cretz (approved):

dandavison:
Thanks for reviewing @cretz!

PR #317, opened 2023-07-28
Require TLS 1.2.

jackdawm:
## What was changed
Added requirement for TLS 1.2 as the minimum version.

## Why?
We should ensure these feature snippets don't allow things the real SDKs disallow.

## Checklist
How was this tested:
I didn't see instructions in the README, so I'm hoping the github workflows run tests.


cretz:
> the real SDKs disallow.

Does the Go SDK disallow this? If so, can't we just rely on that?

PR #329, opened 2023-08-23
Add reset and delete feature test

Sushisource:
## What was changed
Added feature test we'd like to have to verify some server behavior as part of the various pipelines we're adding now.

## Why?
More coverage

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/reset/reset_and_delete/feature.go
+++ b/features/reset/reset_and_delete/feature.go
@@ -1,0 +1,174 @@
+package reset_and_delete
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/api/enums/v1"
+	historyProto "go.temporal.io/api/history/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	SignalName = "drivesignal"
+)
+
+var Feature = harness.Feature{
+	Workflows:   Workflow,
+	Execute:     Execute,
+	CheckResult: CheckResult,
+	// We do this kinda by hand in check result -- it's easier since we want to pass the new run
+	// to check result, but end up looking at the history of the original run
+	CheckHistory: func(ctx context.Context, runner *harness.Runner, run client.WorkflowRun) error {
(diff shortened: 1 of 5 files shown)


features/reset/reset_and_delete/feature.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	err = r.Client.SignalWorkflow(ctx, newRun.GetID(), "", SignalName, "finish")
+	if err != nil {
+		return nil, err
+	}
+
+	return newRun, nil
+}
+
+func CheckResult(ctx context.Context, r *harness.Runner, run client.WorkflowRun) error {
```
cretz, line 88: It is admittedly a bit hard to decide whether to put logic in `Execute` or `CheckResult`, but I think what's here is probably good. Alternatively could move the workflow result check and deletion as part of `Execute`, but meh.

features/reset/reset_and_delete/feature.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	// Ensure original run is findable via visibility & has correct status
+	resp, err := r.Client.ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+		Namespace: r.Namespace,
+		Query:     "WorkflowId = '" + origRun.GetID() + "'",
+	})
+	if err != nil {
+		return err
+	}
+	r.Assert.Equal(1, len(resp.GetExecutions()))
```
cretz, line 138: I personally like `r.Require` instead of `r.Assert` because it fails immediately, but difference is often negligible

cretz (approved):

features/reset/reset_and_delete/feature.go
```diff
@@ -0,0 +1,174 @@
...
+	// Drive it for a bit
+	for i := 0; i < 5; i++ {
+		err := r.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), SignalName, "")
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	time.Sleep(1 * time.Second)
+	// Find event ID of second WFT complete
```
yycptt, line 52: just want to confirm that there's nothing special about the second WFT complete right?

yycptt (approved):

features/reset/reset_and_delete/feature.go
Sushisource: No, nothing special. Just had to pick one.

PR #332, opened 2023-08-25
Don't use vanity import which was breaking external deps

Sushisource:
This was causing problems with `omes` depending on this repo


cretz (commented):
Nice

cretz (approved):

PR #337, opened 2023-09-08
Don't assume we're running inside Docker

josh-berry:
## What was changed

Change the IP used for test summaries to a proper `localhost` IP, because we may not have the 127.x.x.x subnet that is typically present in Docker, and trying to bind to this hard-coded IP fails.

## Why?

Running `temporal-features` fails on macOS outside of Docker because we cannot bind to the IP address 127.7.2.1.

## Checklist

2. How was this tested: Manually

3. Any docs updates needed? No

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	summaryListenAddr = "127.7.2.1:0"
+	summaryListenAddr = "127.0.0.1:0"
 	FeaturePassed     = "PASSED"
 )
 


cretz (approved):

cretz:
Seems to be a couple of failures bit only on docker build. Go sometimes is failing with:

> 2023-09-08T00:40:35.323Z	ERROR	harness/log.go:56	Feature failed	{"Feature": "bugs/go/child_workflow_cancel_panic", "error": "failed replaying current execution: nondeterministic workflow: extra replay command for CancelTimer: (TimerId:5)"}

That may be one that is fixed in newer Go and we need to update docker, unsure. Others are failing with:

> 2023/09/08 00:45:14 failed connecting / describing namespace after 5s, last error: failed reaching server: last connection error: connection error: desc = "error reading server preface: remote error: tls: expired certificate"

Either `TEMPORAL_CLIENT_CERT` and `TEMPORAL_CLIENT_KEY` secrets are no longer valid, or there is an issue with secrets on fork, or there is something else happening here. Worth investigating.

josh-berry:
> > 2023/09/08 00:45:14 failed connecting / describing namespace after 5s, last error: failed reaching server: last connection error: connection error: desc = "error reading server preface: remote error: tls: expired certificate"
> 
> Either `TEMPORAL_CLIENT_CERT` and `TEMPORAL_CLIENT_KEY` secrets are no longer valid, or there is an issue with secrets on fork, or there is something else happening here. Worth investigating.

Replaced the credentials (the cert was expired), and that seems to have fixed the problem.

> > 2023-09-08T00:40:35.323Z	ERROR	harness/log.go:56	Feature failed	{"Feature": "bugs/go/child_workflow_cancel_panic", "error": "failed replaying current execution: nondeterministic workflow: extra replay command for CancelTimer: (TimerId:5)"}
> 
> That may be one that is fixed in newer Go and we need to update docker, unsure.

This one went away after replacing the certs, but I'm suspicious. I'll go ahead and merge this PR since you approved it already, and I can't see how the NDE would be related to this, but will look into updating Go as you suggest.

PR #350, opened 2023-10-20
Cleanup features.go

dandavison:
### What
- Change `features.go` to use a consistent import alias style for all features.

### Why
- It was no longer making sense to allow imports to use the terminal feature name, since there are now several cases where the same terminal feature name is used in multiple different top-level features.

Incidentally this diff was created by GPT4, using the following prompt:

```
Output a new version of this file with all the imports using the same alias style.

Specifically, what that means is the following. Consider these two lines:

update_async_accepted "github.com/temporalio/features/features/update/activities"
...
update_async_accepted.Feature,

That's the style we are going to use for all imports.
Output a new version of the file doing that, and also ensure the imports are sorted 
alphabetically as the comment requests.
Just output the new file, without any extra text or natural language discussion.
```

### DIFF
--- a/features/features.go
+++ b/features/features.go
@@ -1,42 +1,42 @@
 package features
 
 import (
-	"github.com/temporalio/features/features/activity/cancel_try_cancel"
-	"github.com/temporalio/features/features/activity/retry_on_error"
-	"github.com/temporalio/features/features/bugs/go/activity_start_race"
-	"github.com/temporalio/features/features/bugs/go/child_workflow_cancel_panic"
-	activity_on_same_version "github.com/temporalio/features/features/build_id_versioning/activity_and_child_on_correct_version"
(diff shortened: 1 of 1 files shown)


features/signal/basic/README.md
```diff
@@ -0,0 +1,7 @@
+# Signal a workflow
```
dandavison, line 1: Ignore all changes from here onwards; they will disappear when we merge https://github.com/temporalio/features/pull/349

cretz: :+1: Will wait for that merge before approving this one

dandavison: This one's mergeable now.

cretz (approved):

dandavison:
Thanks!

PR #351, opened 2023-10-20
Replace temporalite with the CLI dev-server

tdeebswihart:
## What was changed
Replace uses of temporalite with the official CLI's dev-server

## Why?
Temporalite has been deprecated, and to deprecate gogo/protobuf we need to either update temporalite or the CLI. Since temporalite is deprecated let's just stop using it

## Checklist
1. Closes: N/A
2. How was this tested: I ran the go tests locally: `go run . run --lang go`
3. Any docs updates needed? No


Sushisource (approved):
Thanks!

PR #354, opened 2023-10-24
Misc changes made during Update work

dandavison:
- Rename `update/intercept` feature to `/update/client_interceptor`

- Add a feature demonstrating that activities can be run in a Signal handler, analogous to the `update/activities` feature that demonstrates this for Update handlers.


features/signal/activities/README.md
```diff
@@ -0,0 +1,8 @@
+# Run activities from within a Workflow Signal
+
+Workflows can run activities from within a Signal handler.
+
+# Detailed spec
+
+Workflow Signal handlers are like workflow code in that they can invoke and wait on activities.
+This test invokes a number of activities within the Signal and blocks on them, returning from the Signal when they have all completed.
```
dandavison, line 8: Is this a feature worth having? I created it to test my own understanding, and I couldn't think of a reason why this feature should exist for Update and not Signal and Query (which someone can add later).

cretz: Sure, worth having IMO. Anything that is a clear callout to something we need to support can be a feature.

features/update/client_interceptor/feature.go
cretz, line 1: :+1: On the name change. Not sure why the intercept feature didn't include workflow inbound interceptors.

cretz (approved):

dandavison:
Thanks @cretz.

PR #359, opened 2023-10-26
Replace gogo/protobuf with google/protobuf

tdeebswihart:
**What changed?**

gogo/protobuf has been replaced with Google's official go compiler. 

**Why?**

gogo/protobuf has been deprecated for some time and the community is moving on, building new tools (like vtproto) atop google's v2 compiler.

**How did you test it?**

This branch is actively used by my [SDK](https://github.com/temporalio/sdk-go/pull/1256) and [server](https://github.com/temporalio/temporal/pull/5032) PRs so you can see it in use there

**Potential risks**

None I can think of.

**Breaking changes for developers**

- `*time.Time` in proto structs will now be [timestamppb.Timestamp](https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/types/known/timestamppb#section-documentation)
- `*time.Duration` will now be [durationpb.Duration](https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb)
- V2-generated structs embed locks, so you cannot dereference them. `go vet` will scream at you about this. If you need a copy, use `proto.Clone`.
- Proto objects, or objects embedding protos, cannot be compared using `reflect.DeepEqual` or _anything_ that uses it. This includes `testify` and `mock` equality testers! 
  - You will need to use the `protorequire` ro `protoassert`packages instead. I've implemented proto-compatible assertions there for all cases I've encountered
  - If you need `reflect.DeepEqual` for any reason you can use `go.temporal.io/api/temporalproto.DeepEqual` instead

---
**Release Plan**

Our goal is to release our API, Go API, Go SDK, and Server all at once, then update our UI (and ui-server), CLI, and HTTP API in a second pass. Until the second pass happens no changes to our JSON formatting will be apparent to users.


harness/go/testing/protoassert/assert.go
cretz, line 1: I don't think adding two single-file tiny packages is necessary. Just invoke `Require.True(proto.Equal(a, b))` at the assertion site. If that's not good enough for us later, we can do more advanced things then.

cretz (commented):
> THIS IS A DRAFT PUT UP TO SOLICIT FEEDBACK WHILE THE PROJECT IS IN PROGRESS 

May want to convert the PR to draft.

harness/go/testing/protoassert/assert.go
tdeebswihart: These provide better failure messages; `Require.True(...)` is rather unhelpful when your tests fail

tdeebswihart: Removed.

harness/go/history/fetcher.go
```diff
@@ -98,7 +100,7 @@ func (f *Fetcher) GetExecutions(ctx context.Context) ([]*workflow.WorkflowExecut
 			Namespace:       f.Namespace,
 			MaximumPageSize: 1000,
 			NextPageToken:   nextPageToken,
-			StartTimeFilter: &filter.StartTimeFilter{EarliestTime: &earliest},
+			StartTimeFilter: &filter.StartTimeFilter{EarliestTime: timestamppb.New(earliest)},
```
Quinn-With-Two-Ns, line 103: (More a general gogoproto migration question) Just to confirm this would be a visible change to the API for users correct? and we are OK with  that?

tdeebswihart: To the wire format or to the Go API? The wire format will remain unchanged

Quinn-With-Two-Ns: I meant an API change

tdeebswihart: Ah, yes. It will be and there's no way around it

Quinn-With-Two-Ns (approved):

PR #361, opened 2023-10-27
Misc Python update / signal / child-workflow features

dandavison:


dandavison:
Added all my non-typescript outstanding changes to this PR, so it now contains

- New features `update/basic_async` and `signal/child_workflow`, with Python implementations
- A Python implementation for `update/basic` and `signal/basic`

features/signal/basic/feature.py
```diff
@@ -0,0 +1,30 @@
...
+        return f"{input} {self._state}"
+
+    @workflow.signal
+    async def my_signal(self, arg: str):
+        self._state = arg
+
+
+register_feature(
+    workflows=[Workflow],
+    expect_run_result=SIGNAL_ARG,
```
cretz, line 29: This doesn't seem to match the result. Also, may be missing calls to actually send a signal.

features/signal/child_workflow/feature.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@workflow.defn
+class Workflow:
+    @workflow.run
+    async def run(self) -> str:
+        child_wf = await workflow.start_child_workflow(
+            ChildWorkflow.run, CHILD_WORKFLOW_INPUT
+        )
+        await child_wf.signal("my_signal", SIGNAL_ARG)
```
cretz, line 20:
```suggestion
        await child_wf.signal(ChildWorkflow.my_signal, SIGNAL_ARG)
```

I think the type safe form is better here

features/signal/basic/feature.py
```diff
@@ -0,0 +1,30 @@
...
+
+@workflow.defn
+class Workflow:
+    def __init__(self) -> None:
+        self._state = WORKFLOW_INITIAL_STATE
+
+    @workflow.run
+    async def run(self) -> str:
+        await workflow.wait_condition(lambda: self._state != WORKFLOW_INITIAL_STATE)
+        return f"{input} {self._state}"
```
cretz, line 20: `input` is undefined

features/signal/child_workflow/feature.py
```diff
@@ -0,0 +1,50 @@
...
+        await child_wf
+        return child_wf.result()
```
cretz, line 22:
```suggestion
        return await child_wf
```

features/signal/child_workflow/feature.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from temporalio import workflow
+from temporalio.client import WorkflowHandle
+
+from harness.python.feature import Runner, register_feature
+
+CHILD_WORKFLOW_INPUT = "child-input"
+CHILD_WORKFLOW_INITIAL_STATE = "1"
+SIGNAL_ARG = "2"
```
cretz, line 10: Don't have to change, but for me extract constants out in these tiny tests doesn't have much value

features/signal/child_workflow/feature.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        await child_wf.signal("my_signal", SIGNAL_ARG)
+        await child_wf
+        return child_wf.result()
+
+
+@workflow.defn
+class ChildWorkflow:
+    def __init__(self) -> None:
+        self._state = CHILD_WORKFLOW_INITIAL_STATE
```
cretz, line 28:
```suggestion
        self._state: Optional[str] = None
```

Instead of constants, but this is pedantic, don't have to change

features/update/basic/feature.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @my_update.validator
+    def my_validate(self, arg: str):
+        if arg == BAD_UPDATE_ARG:
+            raise ValueError("Invalid Update argument")
+
+
+async def checker(_: Runner, handle: WorkflowHandle):
+    try:
+        await handle.execute_update("my_update", BAD_UPDATE_ARG)
```
cretz, line 43:
```suggestion
        await handle.execute_update(Workflow.my_update, BAD_UPDATE_ARG)
```

features/update/basic_async/feature.py
```diff
@@ -0,0 +1,60 @@
...
+async def checker(_: Runner, handle: WorkflowHandle):
+    bad_update_handle = await handle.start_update("my_update", BAD_UPDATE_ARG)
+    bad_update_result = UPDATE_RESULT_NOT_YET_RECEIVED
+    try:
+        bad_update_result = await bad_update_handle.result()
+    except WorkflowUpdateFailedError:
+        pass
+    else:
+        assert False, "Expected Update to be rejected due to validation failure"
+    assert bad_update_result == UPDATE_RESULT_NOT_YET_RECEIVED
```
cretz, line 51: Not sure there is value confirming that Python doesn't set a variable when the assignment raises an exception

cretz (commented):
Just minor things, gonna have to wait until update releases of course (hopefully soon)

features/signal/basic/feature.py
dandavison: Oops, I pushed a corrupted version without running all features again. Fixed.

features/signal/child_workflow/feature.py
dandavison: Thanks, agree that's a cleaner form for the initial state. Taken.

features/update/basic_async/feature.py
dandavison: Definitely not! Fixed.

features/signal/child_workflow/feature.py
dandavison: Agree I had too many constants. I think I still like one or two of them where they are repeated several times and critical to test logic.

Sushisource (approved):
Thanks for adding these!

features/signal/child_workflow/feature.py
cretz: May want to apply to other features in this PR

cretz (approved):
LGTM (but has to wait on Python release to merge of course)

dandavison:
Thanks for reviewing. I've converted this to draft as a marker that it can't be merged yet.

PR #362, opened 2023-10-27
Typescript Update features

dandavison:
Typescript implementations of some Update features.

This PR can be used to review the proposed API of Typescript Update support (PR forthcoming). But it cannot be merged until that is merged.

### DIFF
--- a/features/update/activities/feature.ts
+++ b/features/update/activities/feature.ts
@@ -1,0 +1,38 @@
+import { Feature } from '@temporalio/harness';
+import * as wf from '@temporalio/workflow';
+import * as assert from 'assert';
+
+const myUpdate = wf.defineUpdate<void>('myUpdate');
+const activityCount = 5;
+const activityResult = 6;
+
+const activitiesImpl = {
+  async myActivity(): Promise<number> {
+    return activityResult;
+  },
+};
+
+const activities = wf.proxyActivities<typeof activitiesImpl>({
+  startToCloseTimeout: '5s',
+});
+
+export const feature = new Feature({
+  workflow,
+  activities: activitiesImpl,
+  checkResult: async (_, handle) => {
+    await handle.executeUpdate(myUpdate);
+    const result = await handle.result();
+    assert.equal(result, activityResult * activityCount);
+  },
+});
+
+export async function workflow(): Promise<number> {
+  let total = 0;
+  wf.setHandler(myUpdate, async () => {
+    const promises = Array.from({ length: activityCount }, activities.myActivity);
+    const counts = await Promise.all(promises);
+    total = counts.reduce((a, b) => a + b, 0);
+  });
+  await wf.condition(() => total > 0);
+  return total;
+}
--- a/features/update/basic/feature.ts
+++ b/features/update/basic/feature.ts
@@ -3,51 +3,44 @@ import { Feature } from '@temporalio/harness';
 import * as wf from '@temporalio/workflow';
 import * as assert from 'assert';
 
-const WORKFLOW_INITIAL_STATE = '';
-const UPDATE_ARG = 'update-arg';
-const BAD_UPDATE_ARG = 'reject-me';
-const UPDATE_RESULT = 'update-result';
-const UPDATE_RESULT_NOT_YET_RECEIVED = 'update-result-not-yet-received';
-
 const myUpdate = wf.defineUpdate<string, [string]>('myUpdate');
 
 /**
- * A workflow with a signal and signal validator. If accepted, the signal makes
- * a change to workflow state. The workflow does not terminate until such a
- * change occurs.
+ * A workflow with an update and an update validator. If accepted, the update
+ * makes a change to workflow state. The workflow does not terminate until such
+ * a change occurs.
  */
 export async function workflow(): Promise<string> {
-  let state = WORKFLOW_INITIAL_STATE;
-  const handler = async (arg: string) => {
-    state = arg;
-    return UPDATE_RESULT;
-  };
+  let state = '';
   const validator = (arg: string) => {
-    if (arg == BAD_UPDATE_ARG) {
+    if (arg == 'invalid-arg') {
       throw new Error('Invalid Update argument');
     }
   };
+  const handler = async (arg: string) => {
+    state = arg;
+    return 'update-result';
+  };
   wf.setHandler(myUpdate, handler, { validator });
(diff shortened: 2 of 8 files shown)


features/update/basic/feature.ts
```diff
@@ -0,0 +1,53 @@
...
+  const update = (arg: string) => {
+    state = arg;
+    return UPDATE_RESULT;
+  };
+  const validate = (arg: string) => {
+    if (arg == BAD_UPDATE_ARG) {
+      throw new Error('Invalid Update argument');
+    }
+  };
+  wf.setHandler(myUpdate, update, validate);
```
cretz, line 30: Can you confirm the signature of `setHandler` here? I just want to confirm that the update can return the value or a promise of the value.

cretz, line 30:
```suggestion
  wf.setHandler(myUpdate, update, { validate });
```

I wonder if it's a bit clearer for anything after the handler to be an options interface instead of just taking validate as the third param. I don't mind this though, just bring up to get others' opinions.

features/update/activities/feature.ts
```diff
@@ -0,0 +1,38 @@
+import { Feature } from '@temporalio/harness';
+import * as wf from '@temporalio/workflow';
+import * as assert from 'assert';
+
+const myUpdate = wf.defineUpdate<void, any[]>('myUpdate');
```
cretz, line 5: `defineSignal` and `defineQuery` have defaults for their `Args` types params of an empty array so you don't have to specify input types if there are none, should we do the same here? I wonder if we should even have a default of `void` for the `Ret` type since parameterless void updates may be normal.

features/update/activities/feature.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const activities = wf.proxyActivities<typeof activitiesImpl>({
+  startToCloseTimeout: '5s',
+});
+
+export const feature = new Feature({
+  workflow,
+  activities: activitiesImpl,
+  checkResult: async (_, handle) => {
+    await handle.executeUpdate(myUpdate, {});
```
cretz, line 23: We should not require an options object if don't need to set any fields on one

features/update/basic_async/feature.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export async function workflow(): Promise<string> {
+  let state = WORKFLOW_INITIAL_STATE;
+  const update = (arg: string) => {
+    state = arg;
+    return UPDATE_RESULT;
+  };
+  const validate = (arg: string) => {
+    if (arg == BAD_UPDATE_ARG) {
+      throw new Error('Invalid Update argument');
```
cretz, line 27: You don't need to do it in these feature tests, but in the TS tests, confirm that if they use an application error it is serialized properly as the cause of the update failed error client side (allows them to add details)

features/update/basic/feature.ts
dandavison: Confirmed: the handler can be an async function or a normal function. But this was a good call -- when you made this comment the PR did not demonstrate both.

dandavison: Good suggestion. I agree with you and have changed it to use an optional object as last argument instead. I think that's more idiomatic, and much better if we ever need to add additional optional arguments to this function.

features/update/activities/feature.ts
dandavison: Good catch, fixed!

features/update/basic_async/feature.ts
dandavison: Thanks yes my tests in the main PR do this.

features/update/activities/feature.ts
cretz: I am ok requiring an update to specify its return type for two reasons: 1) we can come back later and add a default if we want, and 2) it makes it clear that updates can return if they want. However, I am also ok defaulting it to void for two reasons: 1) people will default to `any` when they don't care, and 2) it will break their TS compilation if they accidentally add a non-void method as a handler without having explicitly specified the expected return type.

So I have no strong opinion here. Would like to hear other team members' opinions, especially those that work w/ this SDK most.

features/update/basic/feature.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as wf from '@temporalio/workflow';
+import * as assert from 'assert';
+
+const WORKFLOW_INITIAL_STATE = '';
+const UPDATE_ARG = 'update-arg';
+const BAD_UPDATE_ARG = 'reject-me';
+const UPDATE_RESULT = 'update-result';
+const UPDATE_RESULT_NOT_YET_RECEIVED = 'update-result-not-yet-received';
+
+const myUpdate = wf.defineUpdate<Promise<string>, [string]>('myUpdate');
```
cretz, line 12: Per https://github.com/temporalio/sdk-typescript/pull/1277, shouldn't this be:

```suggestion
const myUpdate = wf.defineUpdate<string, [string]>('myUpdate');
```

and `Promise` is implied? I wonder if we want to use TS typing tricks to prevent people from explicitly providing `Promse` as a ret type here.

features/update/basic_async/feature.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (arg == 'invalid-arg') {
+      throw new Error('Invalid Update argument');
+    }
+  };
+  wf.setHandler(myUpdate, handler, { validator });
+  await wf.condition(() => state != '');
+  return state;
+}
+
+export const feature = new Feature({
```
mjameswh, line 29: I'm not sure what is the intended scope of this test. Seems like it is testing two distinct scenarios.

Please add a README.md file for this feature.

dandavison: `update/basic_async/README.md` is in `main` (added in a recently-merged PR). I've improved it in this branch.

We're deliberately testing validation in `update/basic` and `update/basic_async`. An alternative would be to have `update/basic`, `update/validation`, `update/basic_async`, and `update/validation_async`. We can move to that if we want. As things stand the non-TS `update/basic/*` are not implementing their spec since they need to test validation.

mjameswh: Don't need to split. As long as the README file clearly define the scope of that test, I don't really mind that much the fact that a single feature test covers two tightly related scenarios.

mjameswh (approved):
LGTM

dandavison:
Thanks @mjameswh!

PR #364, opened 2023-10-30
Use Go SDK dev server and add a couple of .NET update features

cretz:
## What was changed

* Use Go SDK's dev server tooling instead of the built-in one because the built-in one was the inspiration for that one and that one is more maintained (don't want to use built-in dev server when the exact same thing is already in the SDK)
* Update .NET SDK to beta 2
* Add 2 .NET update features (rest will come at later date)


dandavison:
> Use Go SDK's dev server tooling instead of the built-in one

Can you add the reason for the change to the PR description?

cretz:
> Can you add the reason for the change to the PR description?

Added

harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -64,15 +68,15 @@ public Task<WorkflowHandle> StartSingleParameterlessWorkflowAsync()
...
-            {
-                ExecutionTimeout = TimeSpan.FromMinutes(1)
-            });
+        return Client.StartWorkflowAsync(workflow.Name!, Array.Empty<object?>(), NewWorkflowOptions());
     }
 
+    public WorkflowOptions NewWorkflowOptions() =>
+        new(id: $"{PreparedFeature.Dir}-{Guid.NewGuid()}", taskQueue: WorkerOptions.TaskQueue!)
+        {
+            ExecutionTimeout = TimeSpan.FromMinutes(1)
```
dandavison, line 77: Worth making this more aggressive? Presumably we don't actually want anything in this repo taking longer than a few seconds.

features/update/async_accepted/feature.cs
```diff
@@ -0,0 +1,92 @@
...
+        public async Task ShutdownAsync() => shutdown = true;
+
+        [WorkflowSignal]
+        public async Task FinishUpdateAsync() => finishUpdate = true;
+
+        [WorkflowUpdate]
+        public async Task<int> SuccessfulUpdateAsync()
+        {
+            await Workflow.WaitConditionAsync(() => finishUpdate);
+            finishUpdate = false;
```
dandavison, line 32: I'm a bit skeptical about setting this to `false` on exit, since it would affect other in-flight updates. (Obviously it works for the code you have, I'm just bearing in mind that we're partly aiming for this repo to contain good example code for public readers.)

cretz: This workflow does not support multiple in-flight updates. It is normal for people to code their workflows to expected assumptions/constraints. This workflow is specifically for the code that uses it. Many of our feature workflows are developed for certain expectations but not hypothetical other ones. I don't think it's reasonable to review the feature workflows as a general purpose ones.

harness/dotnet/Temporalio.Features.Harness/Runner.cs
cretz: :+1: May be worth suggesting in a more general place for visibility beyond the PR

Quinn-With-Two-Ns (approved):

dandavison (approved):
LGTM! Just some thoughts/suggestions, nothing blocking from me.

Issue #266, opened 2023-04-18
[Feature Request]  Update feature tests


Quinn-With-Two-Ns:
### Is your feature request related to a problem? Please describe.
Add feature tests for update workflow.

- [x] Go
- [x] Java
- [x] Typescript
- [x] Python
- [ ] Dotnet

https://github.com/temporalio/features/tree/main/features/update

josh-berry:
Do we need #150 for this? I noticed that the update tests are failing in CI because the server version is too old.

PR #365, opened 2023-11-01
Python Workflow Update

Sushisource:
DRAFT until Python release (also may be a Core bug to fix)

## What was changed
Add Python workflow update features

## Why?
Yay more languages

## Checklist

Contributes to: https://github.com/temporalio/features/issues/266

2. How was this tested:
IT AM TESTS

3. Any docs updates needed?

### DIFF
--- a/features/update/activities/feature.py
+++ b/features/update/activities/feature.py
@@ -1,0 +1,53 @@
+import asyncio
+from datetime import timedelta
+
+from temporalio import activity, workflow
+from temporalio.client import WorkflowHandle
+
+from harness.python.feature import Runner, register_feature
+
+
+@workflow.defn
+class Workflow:
+    am_done = False
+
+    @workflow.run
+    async def run(self) -> str:
+        await workflow.wait_condition(lambda: self.am_done)
+        return "Hello, World!"
+
+    @workflow.update
+    async def do_activities(self):
+        act_futures = []
+        for i in range(6):
+            act_futures.append(
+                workflow.start_activity(
+                    say_hi, start_to_close_timeout=timedelta(minutes=1)
+                )
+            )
+        results = await asyncio.gather(*act_futures)
+        self.am_done = True
+        return len(results)
+
+
+@activity.defn
+async def say_hi() -> str:
+    return "hi"
+
+
+async def start(runner: Runner) -> WorkflowHandle:
+    await runner.skip_if_update_unsupported()
+    return await runner.start_single_parameterless_workflow()
+
+
+async def check_result(runner: Runner, handle: WorkflowHandle) -> None:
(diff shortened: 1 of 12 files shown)


Sushisource:
Needs https://github.com/temporalio/sdk-python/pull/411 to un-draft

features/update/async_accepted/feature.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @workflow.run
+    async def run(self) -> str:
+        await workflow.wait_condition(lambda: self.am_done)
+        return "Hello, World!"
+
+    @workflow.update
+    async def do_maybe_sleepy_update(self, sleep: bool) -> int:
+        if sleep:
+            await asyncio.sleep(2)
```
cretz, line 23: For my .NET ones, I had this block on signal so that we didn't have to sleep in a workflow in our tests (sleep causes flakes and in general slows the system down)

features/update/activities/feature.py
```diff
@@ -0,0 +1,53 @@
...
+        act_futures = []
+        for i in range(6):
+            act_futures.append(
+                workflow.start_activity(
+                    say_hi, start_to_close_timeout=timedelta(minutes=1)
+                )
+            )
+        results = await asyncio.gather(*act_futures)
+        self.am_done = True
+        return len(results)
```
cretz, line 30: Result not asserted and no result type hint, can probably just remove

features/update/activities/feature.py
cretz, line 1: Haven't brought it up before in the other review and not specific to this review, but arguably a specific "feature" for whether activities work in updates is more of an integration test than a feature. I wouldn't expect an `update/child_workflow` or `update/timer` feature either. I think moving forward we should be pickier with what we consider features.

features/update/deduplication/feature.py
```diff
@@ -0,0 +1,62 @@
...
+    @workflow.run
+    async def run(self) -> str:
+        await workflow.wait_condition(lambda: self.counter == 2)
+        return "Hello, World!"
+
+    @workflow.update
+    async def inc_counter(self):
+        self.counter += 1
+        # Verify that dedupe works pre-update-completion
+        await asyncio.sleep(1)
```
cretz, line 23: Mentioned elsewhere too so I'll stop, but in general would like this to wait on caller stimulus instead of a timer for flake/perf reasons

features/update/async_accepted/feature.py
Sushisource: Sure, totally reasonable, can change them to signals.

features/update/activities/feature.py
Sushisource: True, I'd agree

cretz (approved):

PR #371, opened 2023-11-15
Skip history fetch when unneeded

Sushisource:
## Why?
Failing server tests

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -367,20 +367,25 @@ func (r *Runner) handleSingleHistory(ctx context.Context, client client.Client, 
 		TaskQueue:      feature.TaskQueue,
 		FeatureStarted: r.createTime,
 	}
+	storage := history.Storage{Dir: filepath.Join(r.rootDir, "features", feature.Dir, "history"), Lang: r.config.Lang}
+	// Load all histories from storage to validate against
+	existingSet, err := storage.Load()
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -367,20 +367,25 @@ func (r *Runner) handleSingleHistory(ctx context.Context, client client.Client,
 		TaskQueue:      feature.TaskQueue,
 		FeatureStarted: r.createTime,
 	}
+	storage := history.Storage{Dir: filepath.Join(r.rootDir, "features", feature.Dir, "history"), Lang: r.config.Lang}
+	// Load all histories from storage to validate against
+	existingSet, err := storage.Load()
+	if err != nil {
+		return err
+	}
+	if !r.config.GenerateHistory && len(existingSet.ByVersion) == 0 {
```
cretz, line 376: Technically:

```suggestion
	if !r.config.GenerateHistory && (r.config.DisableHistoryCheck || len(existingSet.ByVersion) == 0) {
```

But this is probably fine too

cretz (approved):

cmd/run.go
Sushisource: Yeah, may as well

Issue #368, opened 2023-11-06
[Feature Request] Add an Eager Workflow Start test for Cloud


antlai-temporal:
### Is your feature request related to a problem? Please describe.

Eager Workflow Start (EWS) is a latency optimization that eliminates matching, and a database round trip, for the first task of a workflow. See #242 for details.  EWS in Cloud is currently blocked on the need of an end-to-end test that demonstrates that the first workflow task is eagerly dispatched to a local worker. 

### Describe the solution you'd like

The solution should reuse the client connection for starter and worker, and initiate a simple workflow with the eager flag set to true. The first task of this workflow should be immediately returned to the starter, and dispatched to the worker without retries.

PR #374, opened 2023-11-18
Add Eager Workflow Start test

antlai-temporal:
Add a test that checks that we can eagerly start a workflow if the namespace 
has enabled it.

Closes 
#368


features/eager_workflow/successful_start/README.md
```diff
@@ -0,0 +1,10 @@
+# Successful start of a workflow using eager mode
+Eager Workflow Start (EWS) is a latency optimization to reduce the time to start processing the first task of a workflow. The starter program provisions a slot in a suitable worker, requests that the server starts a workflow in eager mode, and then, when it receives the first WFT in the response, it directly schedules the task to the worker, eliminating a network roundtrip and a database transaction.
+
+In each scenario, the starter program and the worker should share a client. The starter program will create a simple workflow in eager mode, and then verify that eager mode was actually used, and the first workflow task was processed correctly.
+
+# Detailed spec
+* The `EnableEagerStart` start workflow option should be `true`.
+* The server response to start workflow should include a non-nil `eager_workflow_task` field.
+* The task timeout for the workflow should be large enough to hang the program on a task retry.
```
Quinn-With-Two-Ns, line 9: What is the  goal of large task timeout? If it is just to guard against task retry then why don't you test that directly? You could assert on metrics or add another atomic counter the workflow  function increments.

Quinn-With-Two-Ns:
For Cloud is the positive test case enough? Do we also want to test the negative case? ie when eager start is not requested by the client no task is sent?

features/eager_workflow/successful_start/feature.go
```diff
@@ -0,0 +1,70 @@
...
+		// Instead of failing the test just skip it.
+		msg := fmt.Sprintf("Enable dynamic config system.enableEagerWorkflowStart=true: numEagerlyStarted=%d", numEager)
+		return runner.Skip(msg)
+	}
+	return nil
+}
+
+func EagerDetector(cntEager *atomic.Uint64) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, response interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		err := invoker(ctx, method, req, response, cc, opts...)
```
Quinn-With-Two-Ns, line 56: nit: I would also check the request asked for eager start

features/eager_workflow/successful_start/README.md
antlai-temporal: Need to clarify in the spec better. The problem is that the gprc interceptors only tell us that the server returned the first task, but the worker, even if we had secured a slot beforehand, could still refuse to process it, or may have crashed...

In that exceptional case the task will be retried after the task timeout, and could eventually succeed through the non-eager path. By adding a very large task timeout we effectively disable retries, and we know that success always comes from the eager path. Atomic counter wouldn't help (it only executes once in both cases), and messing with metrics is much more complicated...

features/eager_workflow/successful_start/feature.go
antlai-temporal: Good point, adding the check

features/eager_workflow/successful_start/README.md
Quinn-With-Two-Ns: If the worker crashed then the whole test all crashed so that would be a failure either way. I do see your point though about the worker refusing that's a good reason for the high timeout.

Quinn-With-Two-Ns (approved):

PR #375, opened 2023-11-28
Use the new temporalproto JSON APIs

tdeebswihart:
## What was changed
I bumped the version of the SDK and API to make use of our new temporalproto APIs

## Why?
We made a breaking change in the API


.github/workflows/ci.yaml
```diff
@@ -141,7 +141,7 @@ jobs:
     # TODO: Find some way to automatically upgrade to "latest"
     with:
       do-push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-      go-repo-ref: 'f9d73bfdf7c8d3ec0311306140fbfafa7fb6f9cf'
+      go-repo-ref: '8f2a3c8f967a5898f45df92f6bc42196e4702a2d'
```
tdeebswihart, line 144: This is the first version of the SDK that uses the new APIs

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/temporalio/features/features v1.0.0
 	github.com/temporalio/features/harness/go v1.0.0
 	github.com/urfave/cli/v2 v2.25.7
-	go.temporal.io/sdk v1.25.2-0.20231121222453-f9d73bfdf7c8
+	go.temporal.io/sdk v1.25.2-0.20231129163633-77626eee3015
```
cretz, line 12: Not the biggest fan of using untagged SDKs in the features repo `main` branch. But I guess we had already done it.

tdeebswihart: Yeah. As we discussed (this is for public record, hello!) I'm going to merge from now to keep usage simple since this is a breaking change that hit both the SDK and the Server, but will do so next time.

PR #404, opened 2024-01-12
Gather SDK versions into a single place in ci.yaml

mjameswh:
## What was changed

- In ci.yaml, moved SDK version numbers to a single block, at the top of the file, then reuse those version numbers as variables across the file.

## Why?

- Makes it easier to update version number from an automated script (e.g. using sed or similar).
- Avoid risk of incoherencies (i.e. different version numbers in `build-docker-images` vs `feature-tests-somelang`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,14 +6,6 @@ on: # rebuild any PRs and main branch changes
     branches:
       - main
 
-# Don't add 'v' in front of version numbers
-env:
-  GO_LATEST: '77626eee301574bc4cfd756dbf0641c61e9e6907'
-  TYPESCRIPT_LATEST: '1.9.0'
-  JAVA_LATEST: '1.22.0'
-  PYTHON_LATEST: '1.4.0'
-  CSHARP_LATEST: '0.1.0-beta2'
-
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -11,6 +11,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  latest-sdk-versions:
```
cretz, line 14: Using an entire no-op job just to avoid the variable being used twice is a bit much. Maybe we can just use `env` at the top-level of this file and use `${{ env.NAME_OF_VAR }}` as needed within? I know it may seem side-effecting to alter the environment variables of these runs, but should be harmless (don't name them something too common). That's better than a no-op job IMO.

Sushisource: +1 to that. It's so lame github's parser doesn't allow normal substitutions

mjameswh: That was my initial attempt, but GHA doesn't allow accessing the `env` context from `jobs.<job_id>.with` ([see ref](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability)).

Using repo-level variables would be a possibility, but I don't think we should go there. Unless I'm mistaken, that would make builds non-reproductible.

Sushisource (approved):

PR #410, opened 2024-01-18
Don't run update tests when server may not support update

dandavison:
The server instance used in server CI does not currently support update. Until that is fixed, do not run any update tests without checking for server support.


Sushisource (approved):
We should just implement a proper skip for TS once the fire is out obviously

dandavison:
> We should just implement a proper skip for TS once the fire is out obviously

I'm reverting this PR here: https://github.com/temporalio/features/pull/412

@alexshtin has got update enabled in the server instance used in the feature tests CI, so although a TS skip-if-no-update-support would be nice to have, it is not currently necessary.

PR #413, opened 2024-01-19
Work around java sdk issue 1973

Quinn-With-Two-Ns:
Currently the Java SDK is incorrectly wrapping the `UpdateWorkflowExecption` in a `ExecutionException` on a sync update call if the SDK falls back to polling the update.

work around this issue https://github.com/temporalio/sdk-java/issues/1973


dandavison (approved):
LGTM!

PR #418, opened 2024-01-31
Dotnet Update Features

Sushisource:
## What was changed
Added update features for .net

## Checklist

1. Closes https://github.com/temporalio/features/issues/266

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/update/basic_async/feature.cs
+++ b/features/update/basic_async/feature.cs
@@ -1,0 +1,64 @@
+using Temporalio.Exceptions;
+
+namespace update.basic_async;
+
+using Temporalio.Client;
+using Temporalio.Features.Harness;
+using Temporalio.Worker;
+using Temporalio.Workflows;
+
+class Feature : IFeature
+{
+    [Workflow]
+    class MyWorkflow
+    {
+        private bool shutdown;
+
+        [WorkflowRun]
+        public Task RunAsync() => Workflow.WaitConditionAsync(() => shutdown);
+
+        [WorkflowUpdate]
+        public async Task<string> MyUpdate(string _)
+        {
+            shutdown = true;
+            return "hi";
+        }
+
+        [WorkflowUpdateValidator(nameof(MyUpdate))]
+        public void ValidateMyUpdate(string arg)
+        {
+            if (arg == "invalid")
+            {
+                throw new ApplicationFailureException("invalid");
+            }
+        }
+    }
+
+    public void ConfigureWorker(Runner runner, TemporalWorkerOptions options) =>
+        options.AddWorkflow<MyWorkflow>();
+
+    public async Task<WorkflowHandle?> ExecuteAsync(Runner runner)
+    {
+        await runner.SkipIfUpdateNotSupportedAsync();
+
+        var handle = await runner.Client.StartWorkflowAsync(
+            (MyWorkflow wf) => wf.RunAsync(),
+            runner.NewWorkflowOptions());
+
+        var badUpdateHandle = await handle.StartUpdateAsync(wf => wf.MyUpdate("invalid"));
+
+        try
+        {
+            await badUpdateHandle.GetResultAsync();
+        }
+        catch (WorkflowUpdateFailedException)
+        {
+            // Expected
+        }
+
+        var goodUpdateHandle = await handle.StartUpdateAsync(wf => wf.MyUpdate("valid"));
+        Assert.Equal("hi", await goodUpdateHandle.GetResultAsync());
+
+        return handle;
+    }
+}
--- a/features/update/basic_async/feature.cs
+++ b/features/update/basic_async/feature.cs
@@ -1,7 +1,6 @@
-using Temporalio.Exceptions;
-
 namespace update.basic_async;
 
+using Temporalio.Exceptions;
 using Temporalio.Client;
 using Temporalio.Features.Harness;
 using Temporalio.Worker;
@@ -50,6 +49,7 @@ class Feature : IFeature
         try
         {
             await badUpdateHandle.GetResultAsync();
+            throw new Exception("Expected to fail");
         }
         catch (WorkflowUpdateFailedException)
         {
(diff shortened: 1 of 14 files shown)


harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -152,6 +198,43 @@ public async Task SkipIfAsyncUpdateNotSupportedAsync()
             Client.GetWorkflowHandle("does-not-exist").StartUpdateAsync(
                 "does-not-exist", Array.Empty<object?>()));
 
+    /// <summary>
+    /// Stop the worker.
```
cretz, line 202:
```suggestion
    /// Start the worker.
```

harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -152,6 +198,43 @@ public async Task SkipIfAsyncUpdateNotSupportedAsync()
...
+    /// </summary>
+    public void StartWorker()
```
cretz, line 204: I wonder if there would be value in a single `RestartWorker` call.

harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -43,21 +70,35 @@ public class Runner
...
             }
-            Logger.LogInformation("Checking result of feature {Feature}", PreparedFeature.Dir);
-            await Feature.CheckResultAsync(this, run);
-            await Feature.CheckHistoryAsync(this, run);
-        }, cancellationToken);
+            catch (OperationCanceledException)
+            {
+                // No good option here to not get an exception thrown when waiting on potentially
+                // different worker instances.
+            }
```
cretz, line 100: Might as well move this duplicated exception swallower to inside `StopAndWait`

harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -3,37 +3,64 @@ namespace Temporalio.Features.Harness;
...
+        private readonly CancellationTokenSource workerStopper;
+        private readonly Task workerTask;
+
+        public WorkerState(
+            CancellationTokenSource workerStopper,
+            TemporalWorker worker
+        )
+        {
+            this.workerStopper = workerStopper;
+            workerTask = worker.ExecuteAsync(workerStopper.Token);
```
cretz, line 28:
```suggestion
            workerTask = Task.Run(
                async () =>
                {
                    try
                    {
                        await worker.ExecuteAsync(workerStopper.Token);
                    }
                    catch (Exception e)
                    {
                        // Log e
                        throw;
                    }
                });
```

The existing code is benefitted by the way we wrote the internals of this call that it's not an `async`, but by default in .NET tasks don't start just because you invoke a method that returns a task, so `Task.Run` is best.

Also, technically the problem with the split start/stop approach (and why we avoided it in the .NET worker API) is because you can't see fatal worker exceptions. So that's what the catch here is for. But if we aren't concerned with that, no prob.

cretz (commented):
Only minor things. Overally I'm happy to see how legible the C# logic and workflows are.

harness/dotnet/Temporalio.Features.Harness/Runner.cs
Sushisource: It's not a matter of not being concerned with it - it's that if I need to stop a worker and start a new one at any time, the `RunUntil` approach just doesn't really fit.

In fact it would be really nice to have a way to stop the worker that doesn't force an exception to be thrown. That feels like a bit of an artificial burden.

features/update/basic_async/feature.cs
```diff
@@ -0,0 +1,64 @@
+using Temporalio.Exceptions;
+
+namespace update.basic_async;
```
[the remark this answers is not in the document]
Sushisource: yeah, stupid IDE kept doing this

harness/dotnet/Temporalio.Features.Harness/Runner.cs
Sushisource: Maybe, but, it's a bit less flexible

harness/dotnet/Temporalio.Features.Harness/Runner.cs
Sushisource: This seems to work jut fine without Task.Run too - I thought .NET always eagerly Ran tasks? In any case I put it inside Task.Run but w/o the try/catch since that would log every time it's cancelled and would need the special-case for that too, and the task gets awaited on anyway, so any thrown exception will bubble up.

cretz: > I thought .NET always eagerly Ran tasks?
No, unfortunately not (same with Python), but if you write a method that returns a `Task` instead of an `async` method that returns a task, that runs eagerly. That's why this worked before, because that's what we did internally.

> but w/o the try/catch since that would log every time it's cancelled and would need the special-case for that too, and the task gets awaited on anyway, so any thrown exception will bubble up.

But the exception isn't bubbled up anywhere until `StopAndWait`. But I am fine with that if we are. Also, swallowing a cancel here is easier than swallowing it in your other two places I think.

cretz (approved):

harness/dotnet/Temporalio.Features.Harness/Runner.cs
```diff
@@ -3,37 +3,72 @@ namespace Temporalio.Features.Harness;
...
+        private readonly CancellationTokenSource workerStopper;
+        private readonly Task workerTask;
+
+        public WorkerState(
+            CancellationTokenSource workerStopper,
+            TemporalWorker worker
+        )
+        {
+            this.workerStopper = workerStopper;
+            workerTask = Task.Run(async () => await worker.ExecuteAsync(workerStopper.Token));
```
cretz, line 28:
```suggestion
            workerTask = Task.Run(() => worker.ExecuteAsync(workerStopper.Token));
```

I think this should also work, but it's basically the same thing so no need to change.

Issue #326, opened 2023-08-21
[Feature Request] Unified client gRPC retry periods


lorensr:
### Is your feature request related to a problem? Please describe.

CGS believes all SDKs retry for longer than the longest namespace migration Unavailable period (10s), which is why they say that Cloud's expected downtime is not SDK calls failing—it's just higher latency. It's [possible](https://github.com/temporalio/sdk-typescript/pull/1012) for the TS Client to retry for less than that.

### Describe the solution you'd like

Figure out what default retry period we want Clients to have (& maybe ask Server & Cloud teams what they want Clients to do) and apply it across the board. IMO it should be at least 10s + some buffer.

I think all Workers indefinitely retry retryable error codes, but we could confirm that while we're at it.

### Current status

* TS: 
  * [Client](https://github.com/temporalio/sdk-typescript/blob/8112c9fcd666f093091fc5515b097a8179a59b50/packages/client/src/grpc-retry.ts#L69-L77): maxAttempts 10, factor 2, initial interval 20ms (see [spreadsheet](https://github.com/temporalio/sdk-typescript/pull/1012) with minimum cumulative)
  * Worker: indefinite
* Java:
  * [Client](https://github.com/temporalio/sdk-java/blob/efce2f17cc373cc717cf9152c0d5f59d9969c139/temporal-serviceclient/src/main/java/io/temporal/serviceclient/rpcretry/DefaultStubServiceOperationRpcRetryOptions.java#L34): 1m
  * Worker: indefinite

### Related
https://github.com/temporalio/features/issues/27

lorensr:
Also make sure that Clients retry the initial connection, not just methods called after connection is set up. 

@rross reports that Java Client doesn't retry.

TypeScript does:

<img width="724" alt="image" src="https://github.com/temporalio/features/assets/251288/7dab083a-3405-447a-9e60-127906da60d5">

chronos-tachyon:
NB: the Java retry issue has been filed as a separate ticket (SDK-1742 in internal Jira) that I hope to get to in the next two weeks.  There currently isn't a public GitHub ticket but I'd be happy to create one if it's needed.

PR #435, opened 2024-03-08
Add features tests for gRPC retry behavior

chronos-tachyon:
## What was changed
* Implement a small gRPC proxy in Go that can freeze or fail requests on demand
* Extend features runner to optionally launch the proxy, if the executable can be found
* Add proxy-based retry tests for Go and Java
* Minor style cleanups / refactoring of the features runner

## Why?
We need tests that verify the client's retry behavior.  [A comment on features#326](https://github.com/temporalio/features/issues/326#issuecomment-1732071582) says that @rross reports that sdk-java doesn't retry correctly on the first connection.  Indeed, these new tests confirmed that a bug exists in sdk-java's GrpcRetryer, as the `serverCapabilities.get()` call can throw in a place that the code doesn't anticipate exceptions.

From here, there is some future work to be done to make these tests more generally useful:

1. CI needs to be modified, to build the proxy executable before running the features tests.
2. Test implementations for C#, TypeScript, and Python should be written.


cmd/run.go
```diff
@@ -154,6 +204,22 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
 		return err
 	}
 
+	var fn func(context.Context, *cmd.Run) error
```
Sushisource, line 207:
```suggestion
	var langRunFn func(context.Context, *cmd.Run) error
```

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
+package main
```
Sushisource, line 1: It's not immediately obvious to me why this guy needs to be his own executable? Seemingly the runner could start this up internally rather than needing to start it as a subprocess?

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+  Immediately reject incoming gRPC requests with UNAVAILABLE.
+
+- POST /accept
+  Accept incoming gRPC requests; this is the default.
+
+- POST /freeze
+  Block on incoming accepted gRPC requests.
+
+- POST /thaw
+  Process incoming accepted gRPC requests immediately; this is the default.
```
Sushisource, line 54: I assume posting to these switches to the corresponding "mode"? Might be worth mentioning this modality concept explicitly.

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+	}
+
+	needProxyClose := true
+	defer func() {
+		if needProxyClose {
+			gProxyServer.ForceClose()
+		}
+	}()
+
+	Info("HTTP control server is running on: %s", flagControl)
```
Sushisource, line 139: Nit: If user passes `127.0.0.1:0` this would show 0 rather than the bound port, yes? 

Not really a problem since we're predetermining available ports and using those, but thought I'd mention it since this can be run directly in theory.

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+		return nil
+	}
+	gStateFrozen = false
+	gStateCond.Broadcast()
+	Info("/thaw: proxy is NOT stalling requests")
+	return nil
+}
+
+func HandleHelp(w http.ResponseWriter, r *http.Request) {
+	if path.Clean(r.URL.Path) != "/" {
```
Sushisource, line 303: Nice touch

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+  Block on incoming accepted gRPC requests.
+
+- POST /thaw
+  Process incoming accepted gRPC requests immediately; this is the default.
+`
+
+var ErrUnknownCommand = errors.New("unknown command")
+
+var (
+	flagTrace   bool
```
Sushisource, line 60: A bit in the same vein as asking why this is its own binary, it might be nice to move all of these into some `FeaturesProxyServer` struct or something

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+}
+
+func (s *ProxyServer) finish() {
+	s.mu.Lock()
+	s.gc = nil
+	s.gs = nil
+	s.l = nil
+	s.wc = nil
+	s.ws = nil
+	s.quitCh = nil
```
Sushisource, line 673: Seems like it might be easy to miss any future fields here. I'm wondering if just having three top-level fields: The mutex, condvar, and an "everything else" container might be nice.

It could also potentially let you have some generic code for this pattern that the servers can share.

Sushisource (commented):
This generally makes sense to me and thanks for some of the various cleanup that's in here. My only question is around the need for this proxy to be its own binary/process. We could save a little argparsing and CI code if it's not.

cmd/run.go
cretz, line 207: No disagreement with extracting switch case content into methods, but why move the switch up here? If it's all the same, can just leave switch where it is and make the actual invocations in each case

internal/cmd/temporal-features-test-proxy/main.go
cretz: Concur. @chronos-tachyon - take a peek at what I recently had to do with HTTP proxying: https://github.com/temporalio/features/pull/448 (ignore the subprocess stuff which was because I had to test environment variables safely and unrelated to the HTTP proxy part). Some notes:

* Can have a feature's config opt-in to the proxy so if someone runs a specific feature without needing a proxy it's not started
* Can name it "tcp proxy" to disambiguate from http proxy (note, I called the other "http proxy" to avoid this confusion too)
* Can make a `TCPProxy` utility as a Go library that the command just happens to use. This can help properly encapsulate the tool (even though it'll only ever be used by the cmd) instead of using globals and subprocesses and such
* Features know if they need a proxy, so they can choose to, say, dial to that address instead of the default one. No need to affect existing running features by making them go through the proxy.

cmd/run.go
```diff
@@ -221,6 +287,49 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
 			return err
 		}
 	}
+	r.config.DirectServer = r.config.Server
```
cretz, line 290: IMO the direct server should be the server and the proxy can be provided for features that know they need it (and the features can create clients to the proxy that know they need to test with the proxy)

internal/cmd/temporal-features-test-proxy/main.go
```diff
@@ -0,0 +1,752 @@
...
+			Error("failed to stop HTTP control server: %v", err)
+		}
+	case <-quitCh:
+	}
+}
+
+func (s *ControlServer) serveThread() {
+	defer s.finish()
+
+	err := s.server.Serve(s.l)
```
cretz, line 464: Big fan of using HTTP here. I wish when the "summary server" thing was done HTTP was used there instead of homemade Temporal binary protocol, heh.

internal/cmd/temporal-features-test-proxy/main.go
cretz: When this is made a library as part of a runner, we should let this listener pick the random port and extract the address out of it instead temporary-listener-free-port approach (which is a good approach in general, but won't be needed when this is not a separate process).

cretz (commented):
All makes sense. High-level notes on the separate proxy process.

internal/cmd/temporal-features-test-proxy/main.go
chronos-tachyon: I couldn't do it in-process because, when I did, there were too many CPU-bound goroutines to allow the proxy server's goroutines to get serviced in a timely manner (delays on the order of 20 seconds).

PR #445, opened 2024-04-02
Replace docker-compose with docker compose

tdeebswihart:
## What was changed

`docker-compose` has been replaced with `docker compose`

## Why?

Docker seems to have deprecated it, which is causing build pipelines which use new versions of docker (or the runner image, I didn't investigate) to fail.


bergundy (approved):
Thanks!

PR #447, opened 2024-04-05
HTTP proxy feature

cretz:
## What was changed

* Added HTTP proxy support to the runner (including confirmation
* Added Go HTTP proxy features
  * Added "subprocess" support since Go gRPC's only way of proxying is via env var
  * Skip Go tests when using loopback since that's not supported in Go
* Added Java HTTP proxy features
  * Also fixed issue where Java summary not properly reporting


cretz:
Abandoning in favor of local-repo branch

PR #448, opened 2024-04-05
HTTP proxy feature

cretz:
## What was changed

* Added HTTP proxy support to the runner (including confirmation
* Added Go HTTP proxy features
  * Added "subprocess" support since Go gRPC's only way of proxying is via env var
  * Skip Go tests when using loopback since that's not supported in Go
* Added Java HTTP proxy features
  * Also fixed issue where Java summary not properly reporting


features/client/http_proxy/feature.go
```diff
@@ -0,0 +1,178 @@
...
+
+func Workflow(ctx workflow.Context) (string, error) { return "done", nil }
+
+type HTTPProxyTest struct {
+	UseAuth bool
+}
+
+func (h HTTPProxyTest) Execute(ctx context.Context, r *harness.Runner) (client.WorkflowRun, error) {
+	// Since HTTP proxy in gRPC only works with the environment variable, we have
+	// to test in a subprocess to not infect other things in this process. The
```
Quinn-With-Two-Ns, line 32: Can't we also use a different `Dialer`? Would that be easier then running a separate subprocess?

https://github.com/grpc/grpc-go/blob/master/Documentation/proxy.md

cretz: Using a different dialer means that you have to do the whole `CONNECT` HTTP request part yourself which basically means you're implementing your own client proxy logic. I think we want to prove that the env var works. If there were a better way to just set a boolean like the env var, I'd do that, but I can find no way. The only way the logic in https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go uses a proxy in `mapAddress` is via https://pkg.go.dev/net/http#ProxyFromEnvironment.

Quinn-With-Two-Ns: Hm, but why do we need to test in a subprocess since `HTTPS_PROXY` will only apply if the environment variable matches the URL right? and no other test would connect to the proxies address?

cretz: It doesn't matter the URL you're wanting to connect to, `HTTPS_PROXY` applies for _all_ gRPC client connections. The only URL matching that occurs is checking whether it's `http` vs `https`, but gRPC always assumes `https` (that's why `HTTP_PROXY` doesn't apply, just `HTTPS_PROXY`).

Quinn-With-Two-Ns: Oh OK my mistake I read 
```
The environment values may be either a complete URL or a "host[:port]", in which case the "http" scheme is assumed. The schemes "http", "https", and "socks5" are supported. An error is returned if the value is a different form.

A nil URL and nil error are returned if no proxy is defined in the environment, or a proxy should not be used for the given request, as defined by NO_PROXY.
```
and thought it meant if the URL didn't match it was excluded, I guess it needs to be in `NO_PROXY` which is a lot more annoying.

Quinn-With-Two-Ns (approved):

PR #450, opened 2024-04-11
Fix feature tests for inclusive schedule backfill

dnr:
## What was changed
Fix tests to be aware of change in temporalio/temporal#5179, activated by temporalio/temporal#5698.

## Why?
Inclusive start/end times are more intuitive and expected.

## Checklist

1. Closes 

2. How was this tested:
Fixing existing tests

3. Any docs updates needed?


dnr:
Hmm, it looks like this won't pass here until we release another cli with those changes in the embedded server? I guess for now we should accept both values.

rodrigozhou (approved):

PR #452, opened 2024-04-12
Scrub worker version before checking history

ShahabT:
## What was changed

Updating expected history json.

## Why?

To include the latest event fields added.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/activity/retry_on_error/history/history.java.1.0.6.json
+++ b/features/activity/retry_on_error/history/history.java.1.0.6.json
@@ -106,8 +106,7 @@
             "applicationFailureInfo": {
               "type": "java.lang.IllegalStateException"
             }
-          },
-          "workerVersion": {}
+          }
         }
       },
       {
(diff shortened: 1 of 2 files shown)


cretz:
Two things: 

1. Why is this set to an empty value in history instead of null?
2. You might have more luck setting this value to nil when we "scrub" histories before comparing, so somewhere in https://github.com/temporalio/features/blob/252b8deb93b301d7cea1269c0c7ddfa5533db290/harness/go/history/history.go#L220

ShahabT:
> Why is this set to an empty value in history instead of null?

Server is only [populating](https://github.com/temporalio/temporal/blob/b702f2ab96e8619f136c894576e434ec0c19e499/service/history/api/respondactivitytaskfailed/api.go#L128) the field from the passed RespondActivityTaskFailedRequest.worker_version. So probably SDK sends an empty object?

cretz (approved):

Sushisource (approved):

PR #458, opened 2024-04-16
Fix latest go sdk version to push the right images

dnr:
## What was changed
Switch the go image build back to using a "version" instead of a "repo ref" so that it pushes the right semver-based image tags (including the no-version tag, which is used by server CI).

Also fix the logic in the go sdk build to be consistent with the other languages.

## Why?
Fix server CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
   latest-sdk-versions:
     runs-on: 'ubuntu-latest'
     outputs:
-      go_latest: 'v1.26.1'
+      go_latest: '1.26.1'
       typescript_latest: '1.9.3'
       java_latest: '1.23.0'
       python_latest: '1.4.0'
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Can we make if fail louder if the version string is not in the right format?

dnr:
I realized more after I sent this: it wasn't failing, it was just using that string as a repo ref instead of a version. That's now fixed, I hope

rodrigozhou (approved):

.github/workflows/ci.yaml
```diff
@@ -157,7 +157,7 @@ jobs:
     # TODO: Find some way to automatically upgrade to "latest"
     with:
       do-push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-      go-repo-ref: ${{ needs.latest-sdk-versions.outputs.go_latest }}
+      go-ver: 'v${{ needs.latest-sdk-versions.outputs.go_latest }}'
```
rodrigozhou, line 160: I thought you wanted to remove this `v` too.

dnr: Nope, the all-docker-images.yaml workflow expects (and requires) the `v`: https://github.com/temporalio/features/blob/main/.github/workflows/all-docker-images.yaml#L7-L8
(it looks like it doesn't actually in practice, it might work without, but it does work with, and I'd rather keep things consistent)

PR #459, opened 2024-04-16
Fix build features:go image

rodrigozhou:
## What was changed

Fix build features:go image so it will update the tags correctly, ie., it will make the GHA to update all the tags `features:go`, `features:go-1`, `features:go-1.26`, etc…

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


dnr:
I'm doing this: #458 . This is still treating the go version string inconsistently with the other languages

PR #474, opened 2024-05-07
Add TypeScript HTTP proxy feature

mjameswh:
## What was changed

- Added feature tests for the [HTTP CONNECT proxy feature in TS SDK](https://github.com/temporalio/sdk-typescript/pull/1411).

- Changed the HTTP Proxy URL exposed by the harness to be `http://` rather than `https://`, because a) TS doesn't support HTTPS proxy and b) the simple proxy server we use for testing is not TLS enabled.

- Changed some minor details of the simple proxy server implementation we use for testing:
  - Write the "HTTP/1.1 200" response as a single string (rather than using library-provided helpers), because the later a) causes TCP fragmentation of the response, and b) introduced a `Content-Length: 0` header, both of which were apparently causing issues to NodeJS's HTTP2 implementation.
  - Once the HTTP connection has been downgraded to TCP, wait for the client to send its first network packet before falling into full-duplex mode. It appears that if NodeJS's HTTP2 client implementation receives the server's first HTTP2 packet before it sends its own first HTTP2 packet, the client will terminate the connection (HTTP2 GOAWAY) and fail.


cmd/run.go
```diff
@@ -234,7 +234,7 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
 		if err != nil {
 			return fmt.Errorf("could not start proxy server: %w", err)
 		}
-		r.config.HTTPProxyURL = "https://" + proxyServer.Address
+		r.config.HTTPProxyURL = "http://" + proxyServer.Address
```
cretz, line 237: Hrmm, IIRC for proxy URLs in the ways I've seen them used, usually `http` or `https` signify whether they are used to proxy HTTP or HTTPS connections, not whether they are actually behind TLS themselves. But I may be wrong.

harness/go/harness/httpproxy.go
```diff
@@ -82,21 +82,14 @@ func (h *HTTPConnectProxyServer) handler(w http.ResponseWriter, r *http.Request)
...
-	}
 
 	targetConn, err := net.Dial("tcp", r.URL.Host)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("Upstream conn failed: %v", err), http.StatusBadGateway)
 		return
 	}
 
-	if err := res.Write(clientConn); err != nil {
+	if _, err := clientConn.Write([]byte("HTTP/1.1 200 Connection established\r\n\r\n")); err != nil {
```
cretz, line 92: Hrmm, can you clarify what the problem was with the response written previously? I am just curious.

harness/go/harness/httpproxy.go
```diff
@@ -105,6 +98,20 @@ func (h *HTTPConnectProxyServer) handler(w http.ResponseWriter, r *http.Request)
...
+	// HTTP2 packet, it will terminate the connection (HTTP2 GOAWAY) and fail. Therefore, wait for
+	// the first client packet before falling into full duplex mode.
+	buf := make([]byte, 32 * 1024)
+	readBytes, err := clientConn.Read(buf)
+	if err != nil {
+		panic(fmt.Sprintf("Expected client to send a first packet: %v", err))
+	}
+	if _, err := targetConn.Write(buf[:readBytes]); err != nil {
+		panic(fmt.Sprintf("Sending client's first packet to server failed: %v", err))
+	}
```
cretz, line 113: Interesting, so server was sending something before the client was?

cretz (commented):
Obviously don't want to approve/merge until TS is released w/ the feature

harness/go/harness/httpproxy.go
mjameswh: Two things:
1. Headers were fragmented, each line of header being sent in its own packet. See in the image below packets 13, 15 and 16.
2. It was sending a `Content-Length: 0` header (packet number 15 in the snapshot), which [is forbidden by the HTTP CONNECT spec](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.6) (though the spec also mandates that clients should ignore that header if present).

For some unknown reason, node's HTTP2 implementation (or maybe grpc-js?) was randomly failing the HTTP2 connection soon after establishment (see the `GOAWAY` at packet 37). I don't know exactly why the two factors contributed occurrence of the `GOAWAY` issue (or either both were contributing to it), but I observed that writing the 200 string in a single operation was significantly reducing the frequency of failure. The status string come from [mitmproxy](https://github.com/mitmproxy/mitmproxy), which I used to compare behavior while investigating this issue.

Screenshot 2024-05-08 at 13 20 20

harness/go/harness/httpproxy.go
mjameswh: Exact. In the image above, see packets 19 and 21.

Now, AFAIU, that's technically legit, and it certainly looks like most HTTP2 clients don't stumble on that. But I suppose this would be a rare occurence in a normal, non-proxied HTTP2 session, as the client will send its first HTTP2 packet immediately after completing the TCP handshake (i.e. after the ACK), whereas the server would only send its first packet after receiving that ACK. Even in a TCP Fast Open scenario, the delay between the client's receiving the handshake ACK and receiving the first data packet would very likely be enough for the client-side HTTP2 library to complete its internal initialization and write out its first packet.

Anyway, I'm opening bug tickets in upstream projects regarding this.

harness/go/harness/httpproxy.go
cretz: Interesting, thanks!

features/client/http_proxy_auth/feature.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const feature = new Feature({
+  workflow,
+  async alternateRun(runner) {
+    assert(runner.options.proxyUrl);
+
+    const connection = await Connection.connect({
+      ...runner.connectionOpts,
+      proxy: {
```
cretz, line 17: Per the comment at https://github.com/temporalio/sdk-typescript/pull/1411#discussion_r1593057635, I am afraid this has side effects that affect other features. If environment variables are already supported and that's the primary way to configure, that's good enough, I don't think we need per-client config then (and you'll have to change your test to do what Go did which is make a sub process).

mjameswh:
Reported the RST/code 2/GOAWAY issue upstream: https://github.com/grpc/grpc-node/issues/2744.

cretz (commented):
LGTM. Not wanting to mark approved because don't want to merge (need to leave in branch until next TS release)

cretz (approved):

PR #480, opened 2024-05-15
Add test for non-durable updates shiped on WFT with events

alexshtin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


alexshtin:
I can't recall why I created this one.

PR #484, opened 2024-05-21
Use an explicit server image tag when provided

tdeebswihart:
## What was changed

When running against a provided docker image we now use the _explicit_ image tag I've embedded in our artifacts rather than `latest`.

## Why?

While adding feature tests to our [builds repo](https://github.com/temporalio/docker-builds/pull/200) I discovered that running `latest` doesn't always run the actual image we've loaded. It seems to be running the latest version we've pushed to dockerhub rather than the locally tagged `latest` version.

If this is true (it's what my testing indicates), then we aren't testing the image we think we are in the server's CI. This will fix that, however


dockerfiles/docker-compose.for-server-image.yaml
```diff
@@ -4,7 +4,7 @@ version: '3.5'
 
 services:
   temporal-server:
-    image: temporaliotest/auto-setup:latest
+    image: temporaliotest/auto-setup:${IMAGE_TAG:-latest}
```
Sushisource, line 7: Where does the dash come from?

Sushisource (approved):
Makes sense to me

dockerfiles/docker-compose.for-server-image.yaml
tdeebswihart: https://docs.docker.com/compose/environment-variables/env-file/#interpolation

tdeebswihart:
Hrm, it looks like `uses` won't work like this when called cross-repo: https://github.com/temporalio/docker-builds/actions/runs/9177587129/job/25235699370?pr=200

That's obnoxious

dockerfiles/docker-compose.for-server-image.yaml
Sushisource: Ah, got it. That syntax is, well, it's a syntax.

Sushisource:
> Hrm, it looks like `uses` won't work like this when called cross-repo: https://github.com/temporalio/docker-builds/actions/runs/9177587129/job/25235699370?pr=200
> 
> That's obnoxious

Yah I want to say that's why I didn't do it that way in the first place. GH Actions has some very weird limitations.

tdeebswihart:
> > Hrm, it looks like `uses` won't work like this when called cross-repo: https://github.com/temporalio/docker-builds/actions/runs/9177587129/job/25235699370?pr=200
> > That's obnoxious
> 
> Yah I want to say that's why I didn't do it that way in the first place. GH Actions has some very weird limitations.

Yeah. I've ripped it out with my latest commit. That's deeply unfortunate

Maybe we should rewrite our workflows in Temporal :P

tdeebswihart:
The final action passed (see https://github.com/temporalio/features/actions/runs/9178336653/job/25238013805?pr=484), the above is out of date.

Issue #455, opened 2024-04-15
[Feature Request] Consider TLS handshake failures as worker-fatal errors


cretz:
### Describe the solution you'd like

On any TLS failure do the normal retry we should do for all worker poll calls, then consider it a fatal error to fail the worker. Do not retry TLS handshake issues forever.

### Per-SDK Tickets

- [ ] Go - TODO
- [ ] Java - TODO
- [ ] Core - TODO
- [ ] TypeScript - N/A
- [x] Python - N/A
- [x] .NET - N/A
- [x] PHP - N/A
- [x] Temporal CLI - N/A

PR #490, opened 2024-06-05
Test Go and Java SDK client retry behavior

Quinn-With-Two-Ns:
Test Go and Java SDK client retry behavior:
* Add a gRPC proxy to inject various failure scenarios
* Renamed some stuff to distinguish this proxy from the http proxy used for some other tests
* Test can declare they want the proxy, and if so if they want to use it as their default connection
* Proxy is spun up as part of the harness only if needed and tests would only use it if they ask


cretz (approved):
Nice. I plan on leveraging this for https://github.com/temporalio/features/issues/455 where I'll  have an HTTP endpoint on the control to fail TLS handshakes. So I may end up needing have the same proxy listen on a TLS-enabled listener too that I can tweak the TLS config on. But I don't think it'll be that hard to plug into this existing one (I hope it's just adding a new listener that's https://pkg.go.dev/crypto/tls#NewListener over the existing one).

PR #502, opened 2024-06-25
Update Java SDK

Quinn-With-Two-Ns:
Update Java SDK to v1.24.0


harness/go/history/history.go
```diff
@@ -225,6 +225,8 @@ func scrubRunSpecificScalars(v interface{}) {
 	case *history.ActivityTaskStartedEventAttributes:
 		v.Identity = ""
 		v.RequestId = ""
+		// Disable because cloud may truncate this value
+		v.LastFailure = nil
```
Quinn-With-Two-Ns, line 229: Note, this has nothing to do with upgrading the SDK, this is because Cloud changed a config

Sushisource (approved):

PR #509, opened 2024-07-10
Don't use class attributes for mutable workflow state

dandavison:
We should not use class attributes for mutable workflow state since they are shared between workflow instances within the same worker process. Even if multiple concurrent instances are unlikely in these examples, this repo may be used by users as a source of code examples.


cretz (approved):
_Technically_ the sandbox should keep the state isolated, but I agree it is bad practice (as are all mutable globals in a workflow).

PR #511, opened 2024-07-22
Upgrade Go SDK to 1.28.1

bergundy:


Quinn-With-Two-Ns:
Shouldn't you upgrade the SDK here as well https://github.com/temporalio/features/blob/main/go.mod?

bergundy:
> Shouldn't you upgrade the SDK here as well https://github.com/temporalio/features/blob/main/go.mod?

Done.

Quinn-With-Two-Ns:
I always update these as well, not  sure if it is strictly necessary? 

https://github.com/temporalio/features/blob/main/harness/go/go.mod

https://github.com/temporalio/features/blob/main/features/go.mod

Quinn-With-Two-Ns (approved):

bergundy:
> I always update these as well, not sure if it is strictly necessary?
> 
> https://github.com/temporalio/features/blob/main/harness/go/go.mod
> 
> https://github.com/temporalio/features/blob/main/features/go.mod

Thanks, done.

PR #519, opened 2024-08-06
Try to auto-build python SDK referenced by path

Sushisource:
Needed (well, makes it a lot easier) by omes by-path building

### DIFF
--- a/sdkbuild/python.go
+++ b/sdkbuild/python.go
@@ -92,7 +92,7 @@ func BuildPythonProgram(ctx context.Context, options BuildPythonProgramOptions) 
 			cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
 			err = cmd.Run()
 			if err != nil {
-				return nil, fmt.Errorf("problem installing deps when buildling sdk by path: %w", err)
+				return nil, fmt.Errorf("problem installing deps when building sdk by path: %w", err)
 			}
 			cmd = exec.CommandContext(ctx, "poetry", "build")
 			cmd.Dir = sdkPath
--- a/sdkbuild/python.go
+++ b/sdkbuild/python.go
@@ -74,12 +74,39 @@ func BuildPythonProgram(ctx context.Context, options BuildPythonProgramOptions) 
 	versionStr := strconv.Quote(strings.TrimPrefix(options.Version, "v"))
 	if strings.ContainsAny(options.Version, `/\`) {
 		// We expect a dist/ directory with a single whl file present
-		wheels, err := filepath.Glob(filepath.Join(options.Version, "dist/*.whl"))
+		sdkPath, err := filepath.Abs(options.Version)
+		if err != nil {
+			return nil, fmt.Errorf("unable to make sdk path absolute: %w", err)
+		}
+		triedBuilding := false
+
+	getWheels:
+		wheels, err := filepath.Glob(filepath.Join(sdkPath, "dist/*.whl"))
 		if err != nil {
 			return nil, fmt.Errorf("failed glob wheel lookup: %w", err)
+		} else if len(wheels) == 0 && !triedBuilding {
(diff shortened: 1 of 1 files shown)


sdkbuild/python.go
```diff
@@ -74,12 +74,39 @@ func BuildPythonProgram(ctx context.Context, options BuildPythonProgramOptions)
...
+			}
+			cmd = exec.CommandContext(ctx, "poetry", "build")
+			cmd.Dir = sdkPath
+			cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+			err = cmd.Run()
+			if err != nil {
+				return nil, fmt.Errorf("problem buildling sdk by path: %w", err)
+			}
+			// This constitutes a legitimate use of goto, fight me.
+			goto getWheels
```
dandavison, line 105: Agreed, this use of `goto` is cleaner than anything else I've thought of (and i can imagine some much uglier ways it would often be written). But,

> fight me.

OK! The whole thing can be made even nicer: there are 45 lines of code doing nothing other than getting an absolute path to a Python wheel. Let's put them in a function rather than add all those 45 lines (with their irrelevant local state) to the existing function.

```go
...
	versionStr := strconv.Quote(strings.TrimPrefix(options.Version, "v"))
	if strings.ContainsAny(options.Version, `/\`) {
		wheel, err := getWheel(ctx, options.Version)
		if err != nil {
			return nil, err
		}
		versionStr = "{ path = " + strconv.Quote(wheel) + " }"
	}

...

func getWheel(ctx context.Context, version string) (string, error) {
	// We expect a dist/ directory with a single whl file present
	sdkPath, err := filepath.Abs(version)
	if err != nil {
		return "", fmt.Errorf("unable to make sdk path absolute: %w", err)
	}
	triedBuilding := false

getWheels:
	wheels, err := filepath.Glob(filepath.Join(sdkPath, "dist/*.whl"))
	if err != nil {
		return "", fmt.Errorf("failed glob wheel lookup: %w", err)
	} else if len(wheels) == 0 && !triedBuilding {
		triedBuilding = true
		// Try to build the project
		cmd := exec.CommandContext(ctx, "poetry", "install", "--no-root")
		cmd.Dir = sdkPath
		cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
		err = cmd.Run()
		if err != nil {
			return "", fmt.Errorf("problem installing deps when buildling sdk by path: %w", err)
		}
		cmd = exec.CommandContext(ctx, "poetry", "build")
		cmd.Dir = sdkPath
		cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
		err = cmd.Run()
		if err != nil {
			return "", fmt.Errorf("problem buildling sdk by path: %w", err)
		}
		// This constitutes a legitimate use of goto, fight me.
		goto getWheels
	} else if len(wheels) != 1 {
		return "", fmt.Errorf("expected single dist wheel, found %v - consider cleaning dist dir", wheels)
	}

	absWheel, err := filepath.Abs(wheels[0])
	if err != nil {
		return "", fmt.Errorf("unable to make wheel path absolute: %w", err)
	}
	// There's a strange bug in Poetry or somewhere deeper where, on Windows,
	// the single drive letter has to be capitalized
	if runtime.GOOS == "windows" && absWheel[1] == ':' {
		absWheel = strings.ToUpper(absWheel[:1]) + absWheel[1:]
	}
	return absWheel, nil
}
```

cretz (approved):
I forget why we had originally expected building to be out-of-band and done by user, but this works for me. They still have to install all prerequisites of course, we aren't doing that step automatically.

sdkbuild/python.go
Sushisource: Haha, yes I did imagine this option, but I hate to split out functions that could otherwise be inline.

PR #530, opened 2024-08-27
Automate selection of latest SDK versions

yuandrew:
## What was changed

Add a new `latest-sdk-version` command to query GH API to get the latest release version. Automatically run CI against latest SDK releases.

Also added a mechanism to run the CI manually, specifying a custom SDK version to allow testing of versions that aren't the latest

This currently bumps the following SDK versions:

typescript: 1.10.3 -> 1.11.1
java: 1.25.0 -> 1.25.1
typescript: 1.2.0 -> 1.3.0

## Why?

Prevent losing test coverage for accidentally forgetting to bump a version manually

## Checklist

1. Closes 

2. How was this tested:

Screenshot 2024-08-27 at 3 49 35 PM

Not sure how to test the override mechanism without having this merged into main. Because the newly added `dispatch_workflow` trigger isn't in main, I'm unable to trigger CI manually.

3. Any docs updates needed?


.github/workflows/ci.yaml
```diff
@@ -14,13 +31,64 @@ jobs:
...
+      go_latest: ${{ steps.get_release.go_latest }}
+      typescript_latest: ${{ steps.get_release.typescript_latest }}
+      java_latest: ${{ steps.get_release.java_latest }}
+      python_latest: ${{ steps.get_release.python_latest }}
+      csharp_latest: ${{ steps.get_release.csharp_latest }}
     steps:
-      - run: 'echo noop'
+      - name: Get the latest release version
+        id: get_release
+        run: |
```
cretz, line 42: I wonder, since this is our own Go application here, if we should just make this a command on the features repo. Meaning you could run `go run . latest-sdk-version --lang go` (or maybe show all without `--lang` or maybe some kind of `sdk-info` command and version is just one of the values, or whatever no specific approach required). That way not only this GH workflow can use it, but anyone can use the command (and Go is a better programming language than bash, heh). Sorry I did not think of this earlier.

yuandrew: Good idea, might as well build this into the features tool itself

cretz: Thanks! Yeah, now anyone (including GHA) will be able to get this info

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,70 @@
...
+		sdk = "java"
+	case "ts":
+		sdk = "typescript"
+	case "py":
+		sdk = "python"
+	case "cs":
+		sdk = "dotnet"
+	default:
+		return fmt.Errorf("unrecognized language")
+	}
```
cretz, line 53:
```suggestion
	sdk, err := expandLangName(config.Lang)
	if err != nil {
		return err
	}
```

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,70 @@
...
+	url := fmt.Sprintf("https://api.github.com/repos/temporalio/sdk-%s/releases/latest", sdk)
+	curl := exec.Command("curl", url)
+	out, err := curl.Output()
+	if err != nil {
+		return fmt.Errorf("failed to query the GH API for SDK version: %w", err)
+	}
```
cretz, line 59: This requires curl be installed. Can you change to https://pkg.go.dev/net/http#Get?

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,70 @@
...
+	Lang string
+}
+
+func (p *LatestSdkVersionConfig) flags() []cli.Flag {
+	return []cli.Flag{
+		langFlag(&p.Lang),
+	}
+}
+
+type Version struct {
```
cretz, line 34:
```suggestion
type GitHubApiVersion struct {
```

Just to avoid confusion with the general term for readers, or you can just change the code below to use an anonymous struct like `version := struct { Name string `json:"name"` }{}`.

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,70 @@
...
+		return fmt.Errorf("failed to query the GH API for SDK version: %w", err)
+	}
+	var version Version
+	err = json.Unmarshal(out, &version)
+
+	if err != nil {
+		return fmt.Errorf("failed to decode json response: %w", err)
+	}
+
+	fmt.Println(strings.Replace(version.Name, "v", "", 1))
```
cretz, line 67:
```suggestion
	fmt.Println(strings.TrimPrefix(version.Name, "v"))
```

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (p *LatestSdkVersionConfig) flags() []cli.Flag {
+	return []cli.Flag{
+		langFlag(&p.Lang),
+	}
+}
+
+type Version struct {
+	Name string `json:"name"`
```
cretz, line 35: Are you sure you want release `name` and not `tag_name`? Meh, probably always the same anyways.

yuandrew:
Should be ready for review!

.github/workflows/ci.yaml
```diff
@@ -39,6 +51,59 @@ jobs:
...
+          latest_version=$(./temporal-features latest-sdk-version --lang go)
+          go_latest=$(echo "$latest_version" | sed 's/^v//')
+          if [ "${{ github.event.inputs.go_sdk_version }}" != "" ]; then
+            go_latest="${{ github.event.inputs.go_sdk_version }}"
+            echo "User-provided Go SDK version: $go_latest"
+          else
+            echo "Latest Go SDK release version: $go_latest"
+          fi
+          echo "go_latest=$go_latest" >> $GITHUB_OUTPUT
```
cretz, line 65: May want to default to input, remove the thing removing the `v` from output (you already do that in the cmd), any may not need to log if it was a workflow parameter, e.g.:

```suggestion
          go_latest="${{ github.event.inputs.go_sdk_version }}"
          if [ -z "$value" ]; then
            go_latest=$(./temporal-features latest-sdk-version --lang go --no-v-prefix)
            echo "Derived latest Go SDK release version: $go_latest"
          fi
          echo "go_latest=$go_latest" >> $GITHUB_OUTPUT
```

But not a big deal either way

cretz (approved):
One minor non-blocking suggestion, but LGTM. Would like @Sushisource to look to since he's more familiar with these workflows.

.github/workflows/ci.yaml
yuandrew: I like the suggestions! Cleans up the code a bit

cretz (approved):

.github/workflows/ci.yaml
```diff
@@ -39,6 +51,44 @@ jobs:
 
       - run: go build -o temporal-features
 
+      - name: Get the latest release version
+        id: latest_version
+        run: |
+          go_latest="${{ github.event.inputs.go_sdk_version }}"
+          if [ -z "$go_latest" ]; then
```
Sushisource, line 58: This script is long enough that I'm tempted to say it ought to go in a file, but, I can see how that's _really annoying_ with the amount of variable interpolation going on. So, up to you if you think that's worth it.

I really wish GH's syntax highlighter at least did a better job here.

cmd/latest_sdk_version.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func getLatestSdkVersion(config LatestSdkVersionConfig) error {
+	var sdk string
+	sdk, err := expandLangName(config.Lang)
+	if err != nil {
+		return err
+	}
+
+	url := fmt.Sprintf("https://api.github.com/repos/temporalio/sdk-%s/releases/latest", sdk)
```
Sushisource, line 42: Haha, consistent naming scheme to the rescue!

Sushisource (approved):
This is looking good to me. Nothing blocking. Thanks!

PR #532, opened 2024-09-03
use WorkflowUpdateHandle

stephanos:
## What was changed

Follow up to https://github.com/temporalio/sdk-java/pull/2204

## Why?

Type was renamed.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/update/async_accepted/feature.java
+++ b/features/update/async_accepted/feature.java
@@ -1,9 +1,9 @@
 package update.async_accepted;
 
 import io.temporal.activity.ActivityInterface;
-import io.temporal.client.UpdateHandle;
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns:
You'll want to push this to a separate branch, not main, then point `sdk-java` there. Updating the main branch would break all the other consumers of features until the next java SDK release

stephanos:
Oh, I thought only the Java SDK would run the Java feature tests.

I pushed this to a new branch `WorkflowUpdateHandle`; and will update the Java SDK PR.

PR #485, opened 2024-05-21
PHP SDK integration

roxblnfk:
## What was changed

Added runner preparation:
- [x] configuring SDK version
- [x] downloading PHP vendor packages
- [x] downloading RoadRunner
- [x] do it in a separated directory

Added runner that runs RoadRunner and starts PHP features in a single run.

Implemented features: 
- [x] activity `3 / 5`
- [ ] build_id_versioning `0 / 6` **_(not supported)_**
  - https://github.com/temporalio/sdk-php/issues/311
- [x] child_workflow `3 / 13`
  - https://github.com/temporalio/sdk-php/issues/451
- [x] client `0 / 2`
- [x] continue_as_new `1 / 1`
  - https://github.com/temporalio/sdk-php/issues/453
- [x] data_converter `7 / 7`
  - https://github.com/temporalio/sdk-php/issues/455
  - https://github.com/temporalio/sdk-php/issues/454
- [ ] eager_activity `0 / 1` _(waiting SDK feature)_
  - https://github.com/temporalio/sdk-php/issues/456
- [x] eager_workflow `1 / 1`
- [x] local_activity `0 / 15` _(all cases are empty)_
- [x] query `5 / 5`
- [x] reset `0 / 1` _(not required)_
- [x] schedule `4 / 5` _(our cron doesn't support seconds. **Should it support seconds?**)_
- [x] search_attributes `0 / 2` _(all cases are empty)_
- [x] signal `6 / 7`
  - https://github.com/temporalio/sdk-php/issues/457
- [x] telemetry `0 / 1` _**(OTEL package is not released yet)**_
- [x] timer `0 / 0`
- [x] tracing `0 / 0`
- [x] update `11 / 11`
  - https://github.com/temporalio/sdk-php/issues/458
  - https://github.com/temporalio/sdk-php/issues/463

## Checklist

- Closes #103
- [ ] CI
  - [ ] Register docker image there https://github.com/temporalio/docker-builds/blob/main/.github/workflows/features-integration.yml
    - [x] https://github.com/temporalio/features/pull/541
    - [x] https://github.com/temporalio/features/pull/546
    - [ ] Wait for merging this PR

## How to run

```bash
go build -o temporal-features
./temporal-features run --lang php
```

Windows:
```bash
go build -o temporal-features.exe
./temporal-features.exe run --lang php
```

PR #541, opened 2024-09-23
Add PHP docker image building

roxblnfk:
## What was changed

Added and integrated the `php.Dockerfile` into the build pipeline.

## Why?

To be able to run PHP Harness tests similar to other SDKs, I need to add the `temporaliotest/features:php` image.
Part of https://github.com/temporalio/features/pull/485


Sushisource (commented):
Maybe this should just be included with https://github.com/temporalio/features/pull/485 ?

roxblnfk:
Actually, this was extracted from #485.
The problem is that we can't verify if #485 works in CI until we add the Docker images to the Temporal registry. For this, we will need to make yet another PR to [temporalio/docker-builds](https://github.com/temporalio/docker-builds).

Sushisource (approved):
OK, sure that makes sense

Issue #529, opened 2024-08-21
Centralize dynamic config


Sushisource:
We have some hardcoded dynamic config values for embedded server here: https://github.com/temporalio/features/blob/main/cmd/run.go#L208

As well as a file for the docker way of running things here:
https://github.com/temporalio/features/blob/main/dockerfiles/dynamicconfig/docker.yaml

We should use one file, I suppose keep the yaml format, so that we always use the same options.

Another motivation for this is that the server folks would like to use it so they know what options should apply to server when they're running it in a full infra environment for nightly pipelines.

PR #543, opened 2024-09-24
Centralize dynamic config

Sushisource:
## What was changed
Read dynamic config from a central location

## Why?
Makes it easier for users of the repo in full-on environments like k8s have a file they can just grab and use for server dynamic config.

## Checklist

1. Closes https://github.com/temporalio/features/issues/529

2. How was this tested:
Existing features work that need these features enabled

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,10 @@ Publishing docker images of the features runner/suites is also supported. It may
 [manually](https://github.com/temporalio/features/actions/workflows/all-docker-images.yaml),
 but is also triggered by default on each push to main.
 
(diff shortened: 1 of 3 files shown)


cretz (approved):
Note, one of the reasons we called it `--dynamic-config-value` on the CLI was with the expectation there may be a `--dynamic-config-file` in the future. Would support adding that now if anyone has the time/desire.

Sushisource:
> Note, one of the reasons we called it `--dynamic-config-value` on the CLI was with the expectation there may be a `--dynamic-config-file` in the future. Would support adding that now if anyone has the time/desire.

It requires changes to the server library itself to support stuff properly. Namely, merging of file-based and explicitly passed properties.

PR #552, opened 2024-10-25
Fix python version constraint

dandavison:
The python harness is not compatible with 3.8 since it uses an unsupported dict union syntax. Tightening constraint to `^3.10`.


cretz (approved):

dandavison:
The next release of the Python SDK won't support 3.8.

PR #555, opened 2024-10-25
Update update tests for Java SDK v1.26.1

Quinn-With-Two-Ns:
Test see if head of main fails


Quinn-With-Two-Ns:
Looks like there is some bug auto discovering the Java SDK version, currently just updating the Java SDK version to unblock myself

Sushisource (approved):

PR #556, opened 2024-10-28
automatically pull latest version for build.gradle

yuandrew:
## What was changed

Auto pull the latest release from `sdk-java`

## Why?

This should now match the version that CI pulls from changes made in #530
## Checklist

2. How was this tested:

Pulls the correct version locally

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,33 @@ spotless {
     }
 }
 
+ext {
+    def fetchLatestVersion = {
+        def url = 'https://api.github.com/repos/temporalio/sdk-java/releases/latest'
+        def connection = new URL(url).openConnection()
+        connection.setRequestProperty("Accept", "application/vnd.github.v3+json")
+
+        // Read the entire response as a single string
+        def response = connection.inputStream.text
+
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -25,14 +25,43 @@ spotless {
...
+
 
 dependencies {
     implementation 'ch.qos.logback:logback-classic:1.2.9'
     implementation 'com.google.guava:guava:31.0.1-jre'
     implementation 'com.google.code.gson:gson:2.8.9'
     implementation 'com.jayway.jsonpath:json-path:2.6.0'
     implementation 'info.picocli:picocli:4.6.2'
-    implementation 'io.temporal:temporal-sdk:1.26.1'
+    implementation "io.temporal:temporal-sdk:${latestSdkVersion}"
```
cretz, line 64: I wonder if this is something that needs to be done at runtime of `features` instead of inside the `build.gradle`. IIUC every SDK language in `features` has a fixed default that can be overridden at runtime? Why override in this gradle instead of via CLI args of the `features` CLI?

PR #577, opened 2025-01-14
Fix CI

dandavison:


cretz (approved):

features/update/deduplication/feature.php
```diff
@@ -25,6 +25,7 @@ class FeatureWorkflow
     public function run()
     {
         yield Workflow::await(fn(): bool => $this->counter >= 2);
+        yield Workflow::await(fn(): bool => Workflow::allHandlersFinished());
```
dandavison, line 28: @Sushisource @roxblnfk LGTY?

dandavison: This is addressing the new server change to update errors: when a workflow completes and an update is in Accepted state, that used to be gRPC NOT FOUND, but is now an update failure. Without the change here, we are getting

```
AcceptedUpdateCompletedWorkflow: Workflow Update failed because the Workflow completed before the Update completed.
```

I have to admit I don't fully understand where the gRPC NOT FOUND was being handled, but the fix here seems like good practice and fixes the test.

Sushisource: Yeah I think this change is fine for the purposes of this PR.

Sushisource (approved):

features/update/deduplication/feature.php
roxblnfk: I think we can just suppress this exception: https://github.com/temporalio/sdk-php/commit/0f0ceec6d77e3abdd3e7ca21744e283dc0563476

dandavison: Thanks @roxblnfk. So I guess both work -- I think I'm finding the wait `allHandlersFinished` a bit cleaner since it's something that we recommend. WDYT?

roxblnfk: What about `$this->counter >= 2 && Workflow::allHandlersFinished()`?
Looks like that counter is important

roxblnfk (approved):

dandavison:
There are two remaining problems here:

```
  docker run --rm -i -v /tmp/temporal-certs:/certs features:php-2.12.3 \
  --server $TEMPORAL_CLOUD_ADDRESS \
  --namespace $TEMPORAL_CLOUD_NAMESPACE \
  --client-cert-path /certs/client.pem \
  --client-key-path /certs/client.key
  shell: /usr/bin/bash -e {0}
  env:
    TEMPORAL_CLOUD_ADDRESS: sdk-ci.a2dd6.tmprl.cloud:7233
    TEMPORAL_CLOUD_NAMESPACE: sdk-ci.a2dd6
    REPO_URL: https://github.com/temporalio/features
    FEATURES_BUILT_IMAGE_TAG: features:php-2.12.3
    FEATURES_BUILT_IMAGE_TAGS: features:php-2.12.3;features:php-2.12;features:php-2;features:php
Starting RoadRunner... error
Error starting RoadRunner: handle_serve_command: Function call error:
	serve error from the plugin *rrtemporal.Plugin stopping execution, error: temporal_plugin_serve: failed reaching server: last connection error: connection error: desc = "error reading server preface: read tcp 172.17.0.2:53200->44.241.31.151:7233: read: connection reset by peer"
```

and

```
  docker run --rm -i -v /tmp/temporal-certs:/certs features:go-1.32.1 \
  --server $TEMPORAL_CLOUD_ADDRESS \
  --namespace $TEMPORAL_CLOUD_NAMESPACE \
  --client-cert-path /certs/client.pem \
  --client-key-path /certs/client.key

panic: 
	Error Trace:	/app/features/update/updateutil/history.go:20
	            				/app/features/update/activities/feature.go:49
	            				/app/harness/go/harness/runner.go:120
	            				/app/harness/go/cmd/run.go:271
	            				/app/harness/go/cmd/run.go:230
	            				/app/harness/go/cmd/run.go:246
	            				/app/harness/go/cmd/run.go:38
	            				/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:163
	            				/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:313
	            				/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224
	            				/app/harness/go/cmd/cmd.go:20
	            				/app/prepared/main.go:9
	            				/usr/local/go/src/runtime/proc.go:267
	            				/usr/local/go/src/runtime/asm_amd64.s:1650
	Error:      	Received unexpected error:
	            	after 5s, no workflow(s) found
```

rustatian:
I double checked the configuration ([.rr.yaml](https://github.com/temporalio/features/blob/main/harness/php/.rr.yaml)) and the problem is in TLS I suppose. RR uses the regular, non-encrypted connection, but you're trying to connect using TLS.
You need to update RR's configuration to include these certs (or envs pointing to these certs) as shown here: [link](https://github.com/roadrunner-server/roadrunner/blob/master/.rr.yaml#L244)

dandavison:
Thanks @rustatian! Do you know what recent change might have caused this to break?

rustatian:
Hard to say, might be some update to the Temporal cloud (or instance located by this address `44.241.31.151:7233`) which is now requires a TLS connection only... As far as I know, this is not a SDK-GO update.

cretz (approved):

features/update/deduplication/feature.php
roxblnfk: @dandavison 
There should be a condition `$this->counter >= 2 && Workflow::allHandlersFinished()` because the Workflow might start and finish before the Update methods are called.

PR #584, opened 2025-01-20
updates_do_not_block_continue_as_new

dandavison:
Adds a new feature that confirms the behavior when a client sends an update to a workflow that is continuing-as-new. Specifically, the test guarantees that the update is admitted while the WFT is in-flight. This confirms that:

1. Updates do not block continue as new (unlike signals)
2. An update that is admitted while the workflow is continuing as new lands on the post-CAN workflow run.

Only implemented for Python so far; let's confirm the approach before extending implementation to other languages.


features/continue_as_new/updates_do_not_block_continue_as_new/feature.py
```diff
@@ -0,0 +1,103 @@
...
+        task_queue=runner.task_queue,
+        execution_timeout=timedelta(minutes=1),
+    )
+
+
+async def check_result(runner: Runner, handle: WorkflowHandle) -> None:
+    async def _check_result() -> None:
+        # See docstring at top of file.
+        # Cause an update to be admitted while the first WFT is in progress
+        first_run_wft_is_in_progress.wait()
```
cretz, line 77: This is a thread-blocking call in asyncio. Can you use https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor? Would also request on the in-workflow `wait` but that makes a bit less sense since blocking the workflow event loop is the goal.

features/continue_as_new/updates_do_not_block_continue_as_new/feature.py
```diff
@@ -0,0 +1,103 @@
...
+        # The workflow is now blocking its thread waiting for the update to be admitted
+        update_task = await admitted_update_task(
+            runner.client, handle, Workflow.update, "update-id"
+        )
+        # Unblock the workflow so that it responds to the WFT with a CAN command.
+        update_has_been_admitted.set()
+        # The workflow will now CAN. Wait for the update result
+        update_run_id = await update_task
+        # The update should have been handled on the post-CAN run.
```
cretz, line 86: I think it'd be worth two extra assertions: 1) that the update event is not in the history of the pre-CAN run, and 2) that the update is in the history of the post-CAN run

features/continue_as_new/updates_do_not_block_continue_as_new/feature.py
dandavison: > This is a thread-blocking call in asyncio.

Yes that's deliberate, could you read the docstring at the top of the file and let me know if you agree with the strategy?

cretz:
```suggestion
        asyncio.get_running_loop().run_in_executor(None, first_run_wft_is_in_progress.wait)
```

^ That is the safe way to make a thread-blocking call like this

dandavison: I believe the code is doing this correctly already. This entire coroutine is run in a new event loop in a new thread. See line 93 below:

```python
await asyncio.to_thread(lambda: asyncio.run(_check_result()))
```

cretz: Ah, I see, then I think this can bring up a different issue: our client is not safe for running in separate event loops (because it is not safe for running in separate threads, and each event loop is effectively a thread)

features/continue_as_new/updates_do_not_block_continue_as_new/feature.py
dandavison: Good call, added these.

features/continue_as_new/updates_do_not_block_continue_as_new/feature.py
dandavison: Ah, right. Modified to just do the thread-blocking synchronization itself in a new thread, using `asyncio.to_thread`:
```python
await asyncio.to_thread(first_run_wft_is_in_progress.wait)
```

cretz (approved):
I do fear with single-language features there is no incentive or even reminder to do these in other languages. I wonder if we should track the work anywhere.

dandavison:
> I do fear with single-language features there is no incentive or even reminder to do these in other languages.

It's not clear that this should be a feature test at all, so didn't seem any point in doing it in multiple langauges. If the retries are done by Frontend / History gRPC interceptors / clients as they are supposed to be then, while it's fine to add it as an integration test in SDKs, it's unclear that there's a need for a feature test across SDKs.

> I wonder if we should track the work anywhere.

One possibility would be to set up a nag of some sort that finds feature tests with less than the full complement of SDKs.

PR #586, opened 2025-01-23
Use EventuallyWithT to prevent reset_and_delete feature flakiness

bergundy:


Sushisource (approved):

features/reset/reset_and_delete/feature.go
```diff
@@ -127,18 +125,24 @@ func CheckResult(ctx context.Context, r *harness.Runner, run client.WorkflowRun)
...
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+		assert.Len(t, resp.GetExecutions(), 1)
+		if len(resp.GetExecutions()) != 1 {
+			return
+		}
+		assert.Equal(t, enums.WORKFLOW_EXECUTION_STATUS_TERMINATED, resp.GetExecutions()[0].Status)
+	}, 1*time.Second, 1*time.Minute)
```
cretz, line 144: Can we speed this up, say to:

```suggestion
	}, 200*time.Millisecond, 10*time.Second)
```

cretz (approved):

features/reset/reset_and_delete/feature.go
```diff
@@ -127,18 +125,24 @@ func CheckResult(ctx context.Context, r *harness.Runner, run client.WorkflowRun)
...
+			Namespace: r.Namespace,
+			Query:     "WorkflowId = '" + origRun.GetID() + "'",
+		})
+		assert.NoError(t, err)
+		if err != nil {
+			return
+		}
+		assert.Len(t, resp.GetExecutions(), 1)
+		if len(resp.GetExecutions()) != 1 {
+			return
```
dandavison, line 141: Why do we check this again and why is it OK to return without error here?

dandavison: Oh I think I see -- if it's `!= 1` then the previous `assert.Len` failed and this tick is not clean, and we don't want a panic on the next line. (Would be nice to have a cleaner way!)

dandavison (approved):
Thanks for fixing this.

features/reset/reset_and_delete/feature.go
bergundy: Yeah, EventuallyWithT is notoriously hard to use. `assert` expressions don't break out of the function and `require` breaks the retry loop. So you're stuck with evaluating certain conditions twice.

PR #600, opened 2025-02-17
uv + maturin migration

dandavison:
TODO: incomplete; currently just minimum needed for `sdk-python` CI to pass


cretz (commented):
LGTM, though leaving unapproved while the Python SDK `main` is not yet ready for this change

Issue #227, opened 2023-02-13
[Feature Request] Add samples using our versioning API


Quinn-With-Two-Ns:
### Is your feature request related to a problem? Please describe.

Provide samples in each SDK for using that SDKs workflow versioning API

### Describe the solution you'd like

Code examples showing the three versioning strategy https://community.temporal.io/t/workflow-versioning-strategies/6911

- [ ] Go
- [ ] Java
- [ ] Python
- [ ] Typescript
- [ ] Dotnet

PR #611, opened 2025-03-15
Add deployment version features

antlai-temporal:
## What was changed

Adding Go features for the new versioning API based on Deployments.

1. Closes 
#227


features/deployment_versioning/common.go
cretz, line 1: IMO the generic helpers here should be in the harness and the workflows should be in the feature files themselves (yes even if it means the few lines are copied across them). For the most part we've tried to keep features self contained.

antlai-temporal: I was following `build_id_versioning/common.go`. Coping the workflows everywhere is not an issue (just a few lines), but I wonder if functions like `setCurrent` that are very deployment specific belong to the harness...

cretz: > Coping the workflows everywhere is not an issue (just a few lines)

:+1:

> I wonder if functions like setCurrent that are very deployment specific belong to the harness...

It seems like normal client use just like our users are expected to use. Is this how we expect our users to use the deployment client (i.e. wait until the server reports it at a certain value before making the next call)?

antlai-temporal: Unfortunately, the server API is eventually consistent because we discover worker properties, as opposed to using a declarative approach... This is something I deeply dislike, but after many team discussions It eventually went that way..

In non-test scenarios it may not be that bad, you may only need to check for deployments/versions existence once, since deployments are explicitly deleted and versions are GC slowly. Or use an infinite  loop (k8) to change desired state, ignoring early failures.

Back to the issue, is it ok to pollute handler code with all this stuff only applicable to deployment features?

cretz: > Back to the issue, is it ok to pollute handler code with all this stuff only applicable to deployment features?

I think it's probably ok (so long as well qualified), but if you want to leave here in common that's fine. But I do think the workflows should be in their feature files.

I do think that you as the first client user of versioning are already seeing difficult user experience for programmatically using versioning APIs. I think that may need to be brought up in a devexp context.

PR #617, opened 2025-04-21
Don't ignore errors

dandavison:
DWISOTT


Sushisource (approved):
Thanks Go! I thought the compiler at least warned about un-inspected errors...

PR #625, opened 2025-05-16
Pin protobufjs to 7.5.1

mjameswh:
## What was changed

Pin TS `protobufjs` to 7.5.1.

## Why?

`protobufjs` 7.5.2, released yesterday, appears to be causing some frictions with an internal mechanism of the SDK, which is results in failure of feature tests related to the TS SDK Protobuf Payload Converter feature.

Due to technical constraints of our feature test harness, our feature tests reuse some protobuf message definitions of the SDK itself, rather than defining and compiling their own protobuf messages as one would normally do. SDK's own tests for that same feature, which more closely represent what how a user would normally implement this, are still passing; same for our `protobufs` sample in a TS SDK 1.11.8 + protobufjs 7.5.2 setup.

For those reasons, temporarily silencing this issue seems a reasonable decision. The SDK itself will also pin the `protobufjs` dependency until we can properly investigate and resolve this issue (will be pinned in temporalio/sdk-typescript#1679, issue tracked in in temporalio/sdk-typescript#1717).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,8 @@
         "@temporalio/workflow": "^1.11.1",
         "commander": "^8.3.0",
         "ms": "^3.0.0-canary.1",
-        "proto3-json-serializer": "^1.1.1"
+        "proto3-json-serializer": "^1.1.1",
+        "protobufjs": "7.5.1"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Duh :man_facepalming: I forgot about the way the `^` works in package.json lol

PR #627, opened 2025-05-28
Update graceful worker shutdown

### LINKED ISSUE #261: [Feature Request] Write spec for graceful worker shutdown
Different SDKs implement graceful shutdown differently.

TypeScript traditionally treated the graceful shutdown period as "give up on shutdown after this period" and recently changed the semantics to be inline with Python / .NET where graceful shutdown refers to when to issue cancellations to running activities.

By default all Core based SDKs now will wait for activities to complete before returning from `Worker.run`.

Go does not wait for activities to complete today and should change to be consistent with Core based SDKs.
Need to verify what Java does here.

yuandrew:
## What was changed

Updated existing graceful worker shutdown spec

## Why?

Existing spec was outdated

## Checklist

1. Closes #261 

2. How was this tested:
Tests will be added in a separate Issue/PR, see parent issue for more details

3. Any docs updates needed?


features/activity/shutdown/README.md
```diff
@@ -10,12 +10,15 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
 encounter a hard timeout.
 
 # Detailed spec
-* If a worker is told to begin shutdown, activities are immediately notified via
-  a cancel. It must be possible for the user to determine if this cancel was issued
-  by server, or is the result of worker shutdown
+* When a worker shutdown is initiated, the activity context isn't canceled until shutdown
+  has succeeded or the shutdown timeout has elapsed
```
cretz, line 14: I wonder if we should call this "graceful shutdown timeout" so it's not confused with some actual timeout for shutting down?

features/activity/shutdown/README.md
```diff
@@ -10,12 +10,15 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
...
+* It must be possible for the user to determine if this cancel was issued by server,
```
cretz, line 15: This is only after cancel which means graceful timeout has elapsed. Shouldn't it also be possible for a user to determine that a worker is being shut down even before graceful timeout elapses?

features/activity/shutdown/README.md
cretz, line 14: I don't think "until shutdown succeeded" is accurate here, we cancel _before_ shutdown succeeds (in fact, shutdown can't succeed until all activities complete in Core-based SDKs and maybe others). May just be worth removing:

```suggestion
* When a worker shutdown is initiated, the activity context isn't canceled until
  the shutdown timeout has elapsed
```

cretz, line 14: Do we need to clarify that this shutdown timeout defaults as 0/no-timeout?

yuandrew: yes, good idea to clarify this

features/activity/shutdown/README.md
yuandrew: good point

features/activity/shutdown/README.md
```diff
@@ -10,12 +10,20 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
 encounter a hard timeout.
 
 # Detailed spec
-* If a worker is told to begin shutdown, activities are immediately notified via
-  a cancel. It must be possible for the user to determine if this cancel was issued
-  by server, or is the result of worker shutdown
+* When a worker shutdown is initiated, the activity context isn't canceled until the
+  graceful shutdown timeout has elapsed
+* Graceful shutdown language behavior
+  * Core based SDKs - when graceful shutdown timeout isn't specified, that is treated as no-timeout
```
mjameswh, line 16: You are right that Core defaults to no-timeout.

However, TS actually defaults the graceful shutdown timeout to 0. From a quick look, it seems like Python, .Net and Ruby are all doing the same.

features/activity/shutdown/README.md
```diff
@@ -10,12 +10,20 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
...
+  * Java - there is no timeout parameter set like Go or Core, but instead after calling `shutdown()` 
+    a user can call `workerFactory.awaitTermination(timeout, unit)` and `isTerminated()` to see if the timeout
+    has passed, then it is up to the user to forcibly shutdown or do something else
+* It must be possible for an activity to determine that a worker is being shut down even before graceful timeout elapses
+* It must be possible for the user to determine whether the activity context cancel was the result of worker shutdown,
+  or issued by server
+  * TODO: Java - need to add a way for Activities to know when the worker is being shutdown.
+  * TODO: Typescript - figure out what happens in TS
 * Activities may handle the cancel however they like (including continuing running)
 * Heartbeating is possible while the shutdown process is ongoing
```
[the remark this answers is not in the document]
cretz: Note for Core-based SDKs, heartbeating does not notify about the grace period, that is delivered in a separate manner, and heartbeating continues to work regardless of activity status (during or after grace period) IIUC

features/activity/shutdown/README.md
yuandrew: not quite, if you remember from a few weeks (or months ago), Spencer mentioned when `graceful_shutdown_period` is not set, there is no timeout, so cancels are never issued to activities.

But sounds like from James looking at the core SDKs, each SDK will instead send 0 when the language specific timeout isn't set, meaning no SDK actually has `default == no-timeout`

mjameswh: See [here](https://github.com/temporalio/sdk-core/blob/4d8275431bc92de8ddf62eab9b81e2fbb771b9ea/core/src/worker/activities.rs#L645-L646).

mjameswh: IIRC, we only made that one optional for backward compatibility with TS back then, because the TS Worker was emitting activity cancellation by itself and we didn't want to shake the TS worker's logic at that moment. So we instead made it possible for the TS SDK to disable Core's activity cancellation on shutdown.

TS is now relying on Core for that behavior, just like other SDKs, so I think it would be safe to make that option mandatory in Core's worker config. Definitely not high priority and very low impact, of course.

features/activity/shutdown/README.md
```diff
@@ -10,12 +10,20 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
...
+  * Core based SDKs - when graceful shutdown timeout isn't specified, that is treated as no-timeout
+  * Go - when graceful shutdown timeout isn't specified, this is treated as a 0 second timeout
+  * Java - there is no timeout parameter set like Go or Core, but instead after calling `shutdown()` 
+    a user can call `workerFactory.awaitTermination(timeout, unit)` and `isTerminated()` to see if the timeout
+    has passed, then it is up to the user to forcibly shutdown or do something else
+* It must be possible for an activity to determine that a worker is being shut down even before graceful timeout elapses
+* It must be possible for the user to determine whether the activity context cancel was the result of worker shutdown,
+  or issued by server
+  * TODO: Java - need to add a way for Activities to know when the worker is being shutdown.
+  * TODO: Typescript - figure out what happens in TS
```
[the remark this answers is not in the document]
yuandrew: thank you!

features/activity/shutdown/README.md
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah that is why I put the "may" in there

features/activity/shutdown/README.md
```diff
@@ -10,12 +10,19 @@ complete with failure, and ignore the cancel. The one which ignores the cancel s
...
+* Graceful shutdown language behavior
+  * Core based SDKs - when graceful shutdown timeout isn't specified, that is treated as no-timeout
+  * Go - when graceful shutdown timeout isn't specified, this is treated as a 0 second timeout
+  * Java - there is no timeout parameter set like Go or Core, but instead after calling `shutdown()` 
+    a user can call `workerFactory.awaitTermination(timeout, unit)` and `isTerminated()` to see if the timeout
+    has passed, then it is up to the user to forcibly shutdown or do something else
+* It must be possible for an activity to determine that a worker is being shut down even before graceful timeout elapses
+* It must be possible for the user to determine whether the activity context cancel was the result of worker shutdown,
+  or issued by server
+  * TODO: Java - need to add a way for Activities to know when the worker is being shutdown.
```
yuandrew, line 24: @Quinn-With-Two-Ns  can you remind me, is this still a TODO? I can't remember when/where you last mentioned this.

If it's still true, I can create an GH issue and link the issue to this TODO

PR #628, opened 2025-05-29
Fixed SDK version mismatch in .Net build. Made `--version` optional in `prepare`.

maciejdudko:
## What was changed

- Changed .Net build files so that the tests are built against the same Temporal SDK version they're being run with.
- Made `--version` argument optional in `temporal-features prepare` command.

## Why?

- Previously, .Net harness and tests were built against the hardcoded SDK version but run with the version specified in command line. This would make the code sometimes fail mysteriously due to binary incompatibilities, and caused CI failure in sdk-dotnet PR: https://github.com/temporalio/sdk-dotnet/pull/483/ . Now, the harness and tests are always built and run with the same SDK version.

- `run` command already supported omitting `--version` argument (individual languages still do a check if it's required for them). `prepare` still had it as always required for no real reason, so I changed it to match `run`'s behavior to make testing the other fix easier. This change seems useful in general, so I left it in.

## Checklist

1. How was this tested:

- Before, .Net feature tests failed with MissingMethodException if tested with SDK commit https://github.com/temporalio/sdk-dotnet/pull/483/commits/ab26dfe8337f36be3d9fb698f3826f822248251d . After these changes, the tests pass.

- `temporal-features prepare` with omitted/default version was tested to work correctly with Java, Go and .Net.

### DIFF
--- a/cmd/prepare.go
+++ b/cmd/prepare.go
@@ -44,7 +44,6 @@ func (p *PrepareConfig) flags() []cli.Flag {
 		&cli.StringFlag{
 			Name:        "version",
 			Usage:       "SDK language version to run. Most languages support versions as paths.",
-			Required:    true,
 			Destination: &p.Version,
 		},
 	}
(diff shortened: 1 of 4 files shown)


cretz (approved):
This is great, thanks!

PR #662, opened 2025-08-07
Custom stdout/stderr writer

stephanos:
## What was changed

The SDK builders and runners all log directly to stderr/stdout.

## Why?

When running them from within a Go program, that makes it harder to capture the logs - especially when running in concurrently.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/sdkbuild/dotnet.go
+++ b/sdkbuild/dotnet.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+
+	"io"
 )
 
 // BuildDotNetProgramOptions are options for BuildDotNetProgram.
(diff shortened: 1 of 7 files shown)


sdkbuild/dotnet.go
```diff
@@ -128,12 +130,17 @@ func (d *DotNetProgram) Dir() string { return d.dir }
 
 // NewCommand makes a new command for the given args.
 func (d *DotNetProgram) NewCommand(ctx context.Context, args ...string) (*exec.Cmd, error) {
+	return d.NewCommandWithLogger(ctx, nil, args...)
+}
+
+// NewCommandWithLogger makes a new command for the given args with logger support.
+func (d *DotNetProgram) NewCommandWithLogger(ctx context.Context, logger Logger, args ...string) (*exec.Cmd, error) {
```
stephanos, line 137: I assumed we don't want to break `NewCommand` and created a new function.

sdkbuild/php.go
```diff
@@ -47,7 +49,7 @@ func BuildPhpProgram(ctx context.Context, options BuildPhpProgramOptions) (*PhpP
 
 	// Skip if installed
 	if st, err := os.Stat(filepath.Join(dir, "vendor")); err == nil && st.IsDir() {
-		return &PhpProgram{dir, sourceDir}, nil
+		return &PhpProgram{dir: dir, source: sourceDir}, nil
```
stephanos, line 52: Safer.

Sushisource (approved):

PR #686, opened 2025-10-20
Add TLS Server Name Config

kepe-temporal:
## What was changed
Added TLS Server Name Config
In the harnesses for each language, if the option is set, update the TLS config to override the TLS Server Name to the Passed Config

## Why?
We Want to be able to override the TLS Server Name on the certificate so the client doesn't rely on that in order to connect
This is used internally for conformance testing with a cert that has an invalid server name.

## Checklist
- [ ] Code Looks Good

2. How was this tested:
It was not
The easiest way It seems to test this (once it has been reviewed) is to just publish a new container image of this, then add the parameter to the template for this component, then make sure the infra plane is correctly passing this value (set the value to the current value it is trying to connect to).

### DIFF
--- a/features/deployment_versioning/routing_auto_upgrade/feature.go
+++ b/features/deployment_versioning/routing_auto_upgrade/feature.go
@@ -16,11 +16,11 @@ import (
 var deploymentName = uuid.NewString()
 var v1 = worker.WorkerDeploymentVersion{
 	DeploymentName: deploymentName,
-	BuildId:        "1.0",
+	BuildID:        "1.0",
 }
 var v2 = worker.WorkerDeploymentVersion{
 	DeploymentName: deploymentName,
-	BuildId:        "2.0",
+	BuildID:        "2.0",
 }
 
 func WaitForSignalOne(ctx workflow.Context) (string, error) {
--- a/features/deployment_versioning/routing_pinned/feature.go
+++ b/features/deployment_versioning/routing_pinned/feature.go
@@ -16,11 +16,11 @@ import (
 var deploymentName = uuid.NewString()
 var v1 = worker.WorkerDeploymentVersion{
 	DeploymentName: deploymentName,
-	BuildId:        "1.0",
+	BuildID:        "1.0",
 }
 var v2 = worker.WorkerDeploymentVersion{
 	DeploymentName: deploymentName,
-	BuildId:        "2.0",
+	BuildID:        "2.0",
 }
 
 func WaitForSignalOne(ctx workflow.Context) (string, error) {
--- a/features/deployment_versioning/routing_with_override/feature.go
+++ b/features/deployment_versioning/routing_with_override/feature.go
@@ -16,11 +16,11 @@ import (
 var deploymentName = uuid.NewString()
 var v1 = worker.WorkerDeploymentVersion{
 	DeploymentName: deploymentName,
-	BuildId:        "1.0",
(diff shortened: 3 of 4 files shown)


cmd/run.go
```diff
@@ -92,6 +93,11 @@ func (r *RunConfig) dockerRunFlags() []cli.Flag {
 			Usage:       "Path of TLS client key to use (optional)",
 			Destination: &r.ClientKeyPath,
 		},
+		&cli.StringFlag{
+			Name:        "tls-server-name",
+			Usage:       "TLS server name to use for verification (optional)",
```
cretz:
```suggestion
			Usage:       "TLS server name to use for verification and SNI override (optional)",
```

(pedantic, not a big deal)

harness/go/harness/util.go
```diff
@@ -238,7 +238,11 @@ func LoadTLSConfig(clientCertPath, clientKeyPath string) (*tls.Config, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to load certs: %s", err)
 		}
-		return &tls.Config{Certificates: []tls.Certificate{cert}}, nil
+		tlsConfig := &tls.Config{Certificates: []tls.Certificate{cert}}
+		if tlsServerName != "" {
+			tlsConfig.ServerName = tlsServerName
+		}
```
cretz:
```suggestion
		tlsConfig := &tls.Config{Certificates: []tls.Certificate{cert}, ServerName: tlsServerName}
```

No need for the conditional

harness/dotnet/Temporalio.Features.Harness/App.cs
```diff
@@ -79,19 +84,28 @@ private static async Task RunCommandAsync(InvocationContext ctx)
...
+        if (ctx.ParseResult.GetValueForOption(clientCertPathOption) is { } certPath)
+        {
+            tlsOptions = new()
+            {
+                ClientCert = File.ReadAllBytes(certPath.FullName),
+                ClientPrivateKey = File.ReadAllBytes(
+                    ctx.ParseResult.GetValueForOption(clientKeyPathOption)?.FullName ??
+                    throw new ArgumentException("Missing key with cert"))
+            };
+            if (!string.IsNullOrEmpty(tlsServerName))
```
cretz: Probably don't need this conditional, can just assign to the field (I wouldn't expect it'd come across as empty)

cretz (approved):
Nothing blocking, but can simplify the code a lot by just using the absence/null-ness of the value in the harnesses to just set on the TLS config w/out checking conditionals first.

Let me know if you want to alter that or not, then I can merge whenever ready.

kepe-temporal:
@cretz I updated the sdk version. Seeing some failures in go and typescript, but it appears these have commonly failed, and not related to this pr, but not 100 percent sure.

```
Feature deployment_versioning/routing_auto_upgrade failed: timeout waiting for condition to be met
Feature deployment_versioning/routing_pinned failed: timeout waiting for condition to be met
Feature deployment_versioning/routing_with_override failed: timeout waiting for condition to be met
Feature deployment_versioning/routing_with_ramp failed: timeout waiting for condition to be met
```

kepe-temporal:
Other than that, I think this is good to go for now. Will consider some of the other parts if we use this more (which we should. the oss pipeline uses this, but really we should make this a required part of the cloud release testing), or as another pass if something else breaks.

PR #709, opened 2025-12-11
Remove Autosetup and Add default namespace creation for GHA

chaptersix:
## What was changed
* add back #707
* add default namespace creation

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/dockerfiles/docker-compose.yml
+++ b/dockerfiles/docker-compose.yml
@@ -45,7 +45,7 @@ services:
       restart_policy:
         condition: on-failure
         delay: 5s
-        max_attempts: 3
+        max_attempts: 6
     depends_on:
       cassandra:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


chaptersix:
> Admittedly not doing deep review/scrutiny because we can always adjust things as needed and so long as we get CI to pass here (flakes notwithstanding), we're good

thanks! I'm testing these changes against a server repo PR as well. I'm moving away from a shared compose file due to it being a pita to validate locally and in ci. I don't see a reason to copy part of the chart from the server repo anyway. A change in one of the chart definitions would likely break the other and there's not clear link between the 2 files.

chaptersix:
Validated with 
existing docker-builds: https://github.com/temporalio/temporal/pull/8815
new server repo build process: https://github.com/chaptersix/temporal/pull/5

PR #746, opened 2026-02-03
Don't pass AllowNoPollers=true when unsetting current and ramping version

carlydf:
## What was changed
Don't pass AllowNoPollers=true when unsetting current and ramping version.
And remove unnecessary code that used deprecated version string field.

## Why?
Setting AllowNoPollers=true causes the Deployment Client to call `update-with-start` instead of just `update`. In that code path, we verify that Build ID is not empty, because setting a version as current with no pollers creates a new version.

This requires a bug fix in the server, but I don't think it needs to be a patch because it's unlikely people will hit this I think?

## Checklist

1. Closes 

2. How was this tested:
CI

3. Any docs updates needed?

### DIFF
--- a/features/deployment_versioning/aa_clean_old_deployments/feature.go
+++ b/features/deployment_versioning/aa_clean_old_deployments/feature.go
@@ -23,7 +23,10 @@ var Feature = harness.Feature{
 		return run, nil
 	},
 	StartWorkflowOptionsMutator: func(o *client.StartWorkflowOptions) {
-		o.WorkflowExecutionTimeout = 0
+		// I observed this complete in 2 minutes 55 seconds when there were many Deployments
+		// to delete. If the deletions are happening frequently, this will likely be shorter.
+		// Giving a long timeout here to make sure it doesn't cause flakiness.
+		o.WorkflowExecutionTimeout = 5 * time.Minute
 	},
 }
 


features/deployment_versioning/aa_clean_old_deployments/feature.go
```diff
@@ -60,10 +60,6 @@ func ListOldDeployments(ctx context.Context) ([]string, error) {
 			allDeployments = append(allDeployments, deployment.Name)
 		}
 	}
-
-	if err != nil {
-		return nil, err
-	}
```
carlydf: err was always nil here, so this was not doing anything

features/deployment_versioning/aa_clean_old_deployments/feature.go
```diff
@@ -22,6 +22,12 @@ var Feature = harness.Feature{
 		}
 		return run, nil
 	},
+	StartWorkflowOptionsMutator: func(o *client.StartWorkflowOptions) {
+		// I observed this complete in 2 minutes 55 seconds when there were many Deployments
+		// to delete. If the deletions are happening frequently, this will likely be shorter.
+		// Giving a long timeout here to make sure it doesn't cause flakiness.
+		o.WorkflowExecutionTimeout = 5 * time.Minute
```
carlydf: the next run completed in 279ms, so I think when there isn't much work to be done, this workflow completes quickly, but when there are many versions to delete, it takes a while. I think we should keep it at 5 minutes

PR #748, opened 2026-02-06
Wait for RoutingConfigPropagation to complete to fix failing/flaky tests

Shivs11:
## What was changed
- WISOTT
- Moreover, also had to upgrade the API so that I can make gRPC calls. The reason why I am making gRPC calls and not using the SDK is because the released SDK does not have this proto.

## Why?
- to fix breaking things

## Checklist

1. Closes 

2. How was this tested:
- Ran this locally, no failures.

3. Any docs updates needed?

### DIFF
--- a/features/deployment_versioning/common.go
+++ b/features/deployment_versioning/common.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/temporalio/features/harness/go/harness"
 	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/worker"
 	"go.temporal.io/sdk/workflow"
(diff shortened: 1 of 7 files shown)


features/deployment_versioning/common.go
```diff
@@ -112,10 +116,30 @@ func SetRamp(r *harness.Runner, ctx context.Context, deploymentName string, vers
 	_, err = dHandle.SetRampingVersion(ctx, client.WorkerDeploymentSetRampingVersionOptions{
 		BuildID:       version.BuildID,
 		ConflictToken: response1.ConflictToken,
-		Percentage:    float32(100.0),
+		Percentage:    percentage,
```
Shivs11: think this was just an unrelated bug that I noticed......we were passing in percentage but never really using it

kepe-temporal (approved):
LGTM

PR #750, opened 2026-02-10
activity/shutdown: Increase timeout for ScheduleToClose for succeeding workflow

yuandrew:
## What was changed

Create separate timeout `30s` for the successing/failing workflows, while keeping the `ignore` workflow at 300ms. This should not increase the time it takes for tests to finish, but should reduce the raciness, if `cancellSuccess/cancelFailure` couldn't ocmplete within 300ms

Change TS to use setTimeout to not react to cancellation. 

## Why?

Reduce test flakes

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/activity/shutdown/feature.cs
+++ b/features/activity/shutdown/feature.cs
@@ -45,7 +45,7 @@ class Feature : IFeature
             };
             var ignoreOptions = new ActivityOptions
             {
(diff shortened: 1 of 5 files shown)


mjameswh (approved):
LGTM

PR #758, opened 2026-02-18
Update TS SDK to 1.15.0

mjameswh:
## What changed

- Update TS SDK to 1.15.0
- Update Node, TypeScript, eslint, and more, to match now-preferred TS SDK's setup

### DIFF
--- a/features/activity/shutdown/feature.ts
+++ b/features/activity/shutdown/feature.ts
@@ -47,8 +47,8 @@ export async function workflow(): Promise<string> {
   const fut1 = gracefulActivities.cancelFailure();
   const fut2 = ignoringActivities.cancelIgnore();
   // Register rejection handlers eagerly in case harness is slow seeing first activity scheduled event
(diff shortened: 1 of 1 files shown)


features.sln
```diff
@@ -0,0 +1,24 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
Sushisource: Shouldn't get checked in

package.json
```diff
@@ -32,6 +32,6 @@
     "prettier": "^2.3.2",
     "ts-node": "^10.4.0",
     "tsconfig-paths": "^3.12.0",
-    "typescript": "^5.8.3"
+    "typescript": "^5.9.3"
```
chris-olszewski: I think this bump is causing the typecheck failures due to the changes to the `Buffer` family of types in 5.9

chris-olszewski (approved):

PR #768, opened 2026-03-08
Resolve latest SDK versions from package registries

### LINKED ISSUE #761: Have features repo pull the package manager's latest SDK version
### Is your feature request related to a problem? Please describe.

We are using lock files and specific versions for running features locally (CI automatically pulls latest via GH API)

### Describe the solution you'd like

Allow features to default to the latest SDK version, which each lang's package manager should hopefully automatically do if we remove the set version currently.

THardy98:
## Summary
- Query per-language package registries (npm, PyPI, Packagist, Maven Central, NuGet, RubyGems, Go proxy) instead of GitHub API tags to resolve latest SDK versions — avoids cases where a tag exists but the package was never published (e.g. PHP 2.17.1)
- Restore lock files and use targeted SDK update commands (`npm update`, `uv lock --upgrade-package`) in CI so non-SDK deps stay pinned
- Fix PHP feature tests: replace `Temporal\Testing\Environment` with direct `Process` usage for RoadRunner lifecycle — the Environment class's pre-flight check assumed it started the Temporal server, but the Go harness starts it externally
- Bump mypy to `>=1.0` for compatibility with latest protobuf stubs

Closes #761

### DIFF
--- a/features/activity/cancel_try_cancel/feature.cs
+++ b/features/activity/cancel_try_cancel/feature.cs
@@ -86,7 +86,7 @@ class Feature : IFeature
             }
 
             // Send result as signal to workflow
-            await client.GetWorkflowHandle<MyWorkflow>(ActivityExecutionContext.Current.Info.WorkflowId).
+            await client.GetWorkflowHandle<MyWorkflow>(ActivityExecutionContext.Current.Info.WorkflowId!).
                 SignalAsync(wf => wf.SetActivityResultAsync(result));
         }
     }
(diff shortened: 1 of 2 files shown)


Sushisource (commented):
Hmm... I get why we removed the lock files to allow this to work, but, ideally we want to be able to keep the various _other_ deps locked still, since otherwise we're subject to random upgrade-related issues.

We could have the harness run the npm/uv update before executing?

PR #798, opened 2026-04-07
docs: add Python external storage snippets

lennessyy:
Add snipsync snippets for the External Storage Python SDK documentation, including S3 driver setup and custom storage driver examples.

## What was changed

- Added code samples for external storage in Python, to be synced to docs

## Why?

For better validated code samples in docs. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


features/snippets/external_storage/custom_driver_data_converter.py
```diff
@@ -0,0 +1,17 @@
+import dataclasses
```
jmaeagle99: Do we want each of these in their own separate folders since we'll likely have 6-8 of these (one for each language) eventually? Separate folders by concept, not by language. See how plugins was done: https://github.com/temporalio/features/tree/main/features/snippets/plugins

jmaeagle99 (approved):

PR #809, opened 2026-04-27
Specify additional dependencies with Ruby build

THardy98:
## What was changed
DWISOTT

## Why?
Allows support of arbitrary dependencies, including local dependencies.

### DIFF
--- a/sdkbuild/ruby.go
+++ b/sdkbuild/ruby.go
@@ -108,15 +108,14 @@ func BuildRubyProgram(ctx context.Context, options BuildRubyProgramOptions) (*Ru
 	} else if options.Version != "" {
 		version := strings.TrimPrefix(options.Version, "v")
 		gemfileLines = append(gemfileLines, fmt.Sprintf(`gem "temporalio", %q`, version))
-	} else {
-		// No version constraint — keep relying on the source gemspec.
 	}
+
 	moreDependencyLines, err := renderRubyDependencies(options.MoreDependencies)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


sdkbuild/ruby.go
```diff
@@ -95,25 +104,22 @@ func BuildRubyProgram(ctx context.Context, options BuildRubyProgramOptions) (*Ru
...
+		gemfileLines = append(gemfileLines, fmt.Sprintf(`gem "temporalio", path: %q`, gemPath))
 	} else if options.Version != "" {
 		version := strings.TrimPrefix(options.Version, "v")
-		gemfileContent = fmt.Sprintf(`source "https://rubygems.org"
-
-gem "temporalio", "%s"
-gemspec path: %q
-`, version, sourceDir)
-	} else {
-		// No version constraint — Bundler resolves to latest from RubyGems
```
chris-olszewski, line 111: Was dropping support for `temporalio` without a version constraint intetional?

chris-olszewski (approved):

sdkbuild/ruby.go
THardy98: Ah, so we're not dropping support for this, it's just that the code to do this is no longer in the conditional:
```
	gemfileLines := []string{`source "https://rubygems.org"`, ""}
	... branch conditionals where `version` is specified...
	gemfileLines = append(gemfileLines, fmt.Sprintf(`gemspec path: %q`, sourceDir), "")
```
because we did the same thing regardless of which conditional branch we went down

PR #816, opened 2026-05-13
Add initial Nexus support

Quinn-With-Two-Ns:
## What was changed

Add initial Nexus support. Just a basic test in Go and Java will follow up with more languages and test

## Why?

Catch any regressions with Nexus support early, track cross SDK consistency. Same reason we add any feature here.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -649,8 +649,11 @@ func (r *Runner) createNexusEndpoints(ctx context.Context, run *cmd.Run) (func()
 		cl.Close()
 	}
 
+	// Nexus endpoint names have stricter, DNS-style validation than task queues, so / and _
+	// in the feature dir must be normalized to -.
+	sanitize := strings.NewReplacer("/", "-", "_", "-")
 	for _, feature := range nexusFeatures {
-		name := "features-nexus-" + feature.TaskQueue
+		name := "features-nexus-" + sanitize.Replace(feature.Dir) + "-" + uuid.NewString()
 		createCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 		res, err := cl.OperatorService().CreateNexusEndpoint(createCtx, &operatorservice.CreateNexusEndpointRequest{
 			Spec: &nexuspb.EndpointSpec{
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -23,11 +23,16 @@ import (
 	"github.com/temporalio/features/harness/go/history"
 	"github.com/temporalio/features/sdkbuild"
 	"github.com/urfave/cli/v2"
+	nexuspb "go.temporal.io/api/nexus/v1"
+	"go.temporal.io/api/operatorservice/v1"
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/log"
 	"go.temporal.io/sdk/testsuite"
 	"gopkg.in/yaml.v3"
 )
+
+// nexusFeatureDirPrefix marks features that require a per-test Nexus endpoint.
+const nexusFeatureDirPrefix = "nexus/"
 
 const (
 	summaryListenAddr = "127.0.0.1:0"
@@ -267,6 +272,15 @@ func (r *Runner) Run(ctx context.Context, patterns []string) error {
 		defer proxyServer.Close()
 	}
 
+	// Create a Nexus endpoint per feature under features/nexus/ targeting that feature's task
+	// queue. Endpoint names are passed to the lang harness through RunFeature.NexusEndpoint and
+	// the endpoints are deleted once the lang harness completes.
+	deleteEndpoints, err := r.createNexusEndpoints(ctx, run)
+	if err != nil {
+		return err
+	}
+	defer deleteEndpoints()
+
 	// Ensure any created temp dir is cleaned on ctrl-c or normal exit
 	if r.config.DirName == "" && !r.config.RetainTempDir {
 		c := make(chan os.Signal, 1)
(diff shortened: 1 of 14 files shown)


cmd/run.go
```diff
@@ -588,6 +602,80 @@ func rootDir() string {
...
+			_, err := cl.OperatorService().DeleteNexusEndpoint(delCtx, &operatorservice.DeleteNexusEndpointRequest{
+				Id:      ep.ID,
+				Version: ep.Version,
+			})
+			cancel()
+			if err != nil {
+				r.log.Warn("Failed deleting nexus endpoint", "Endpoint", ep.Name, "Error", err)
+			}
+		}
+		cl.Close()
```
bergundy, line 649: Want to also log if close fails?

cmd/run.go
```diff
@@ -588,6 +602,80 @@ func rootDir() string {
...
+
+	type createdEndpoint struct {
+		ID      string
+		Version int64
+		Name    string
+	}
+	var created []createdEndpoint
+	cleanup := func() {
+		for _, ep := range created {
+			delCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
bergundy, line 639: Should this function get a context passed in so it respects any outer cancelations?

cmd/run.go
```diff
@@ -588,6 +602,80 @@ func rootDir() string {
...
+			cancel()
+			if err != nil {
+				r.log.Warn("Failed deleting nexus endpoint", "Endpoint", ep.Name, "Error", err)
+			}
+		}
+		cl.Close()
+	}
+
+	for _, feature := range nexusFeatures {
+		name := "features-nexus-" + feature.TaskQueue
```
bergundy: Note that endpoint names have stricter validation than task queues (because we think we may want to use them as DNS entries at some point).

features/nexus/sync_success/README.md
```diff
@@ -0,0 +1,9 @@
+# Nexus sync operation succeeds
+
+A workflow invokes a synchronous Nexus operation and receives its result.
+
+# Detailed spec
```
bergundy, line 5: Worth checking that the operation transitions to completed state, skipping started in this case.

bergundy (approved):
Worth calling out that this doesn't work in cloud while (I think?) the rest of the features do. It's possible to make it work but endpoint creation is much slower in cloud and you'll need admin credentials to the control plane.

I'd start by documenting that this is OSS only.

Feel free to merge after addressing any meaningful comments.

cmd/run.go
Quinn-With-Two-Ns: The caller has not context to pass in so don't think it is worth it at this point. The caller would just pass in `context.Background()`

cmd/run.go
Quinn-With-Two-Ns: Close can't fail, it doesn't return an error. If the Go SDK does detect an issue it will log though

features/nexus/sync_success/feature.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"github.com/temporalio/features/harness/go/harness"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+)
+
+const ServiceName = "test-service"
+
+var SayHelloOperation = nexus.NewSyncOperation(
```
chrsmith: Minor suggestion: Since we don't care about the actual operation's logic, do you think it would be net clearer to name this `NexusSyncOperation` instead? i.e. describe the _kind_ of thing it is, since "SayHello" isn't that important in this context?

harness/go/cmd/run.go
```diff
@@ -50,18 +50,25 @@ func (r *Run) ToArgs() []string {
...
+			ret[i] += ":" + feature.NexusEndpoint
+		}
 	}
 	return ret
 }
 
 // FromArgs converts the given arguments to features to run.
 func (r *Run) FromArgs(args []string) error {
 	for _, arg := range args {
-		colonIndex := strings.Index(arg, ":")
```
chrsmith, line 60: Nit: For readability, what do you think about moving this parsing logic into a separate function? Something like:

```go
feature, err := parseRunFeature(arg)
if err != nil {
   return fmt.Errorf("parsing argument %q: %w", arg, err)
}
r.Features = append(r.Features, feature)
```

chrsmith (approved):
A couple of nits, but LGTM.

features/nexus/sync_success/feature.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		} else if !ok {
+			return fmt.Errorf("did not find NexusOperationCompleted event in history")
+		}
+		if ok, err := hasEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_STARTED); err != nil {
+			return err
+		} else if ok {
+			return fmt.Errorf("unexpected NexusOperationStarted event for sync operation")
+		}
+		return nil
```
yuandrew, line 77: Do we also want an assertion for `enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED`? The README mentions scheduled but we don't assert for it here

> A sync operation never enters the started state — it transitions directly from scheduled to completed.

yuandrew: same for java

yuandrew (approved):

PR #838, opened 2026-05-28
Add support for testing different Dynamic Configs, add shutdown tests

### LINKED ISSUE #835: Shutdown Worker features tests, enableCancelWorkerPollsOnShutdown
There is new worker shutdown behavior between server and SDK, We should make sure we have tests covering the old behavior, `frontend.enableCancelWorkerPollsOnShutdown=false`, for when this DC gets eventually turned on by default.

yuandrew:
## What was changed
Added support for testing different Dynamic Config settings via existing .config.json, and added tests for `frontend.enableCancelWorkerPollsOnShutdown` and shutdown.

From `.config.json`, you can specify both the dynamic configs to set, as well as `expectNamespaceCapabilities` to assert that the dynamic config was actually set for each run.
  - Added `runVariants` support to the runner.
  - Each variant runs in its own embedded dev server, namespace, and task queue.
  - Variant `dynamicConfig` values override the repo base dynamic config from `dockerfiles/dynamicconfig/docker.yaml`.
  - Added `expectNamespaceCapabilities` validation via `DescribeNamespace` before each variant runs.
  - Exposed validated capability expectations to feature code through `FEATURE_NAMESPACE_CAPABILITIES` for variant-
  specific assertions.

The flow I'd expect for a new feature:

1. Write feature, merge to main
2. Write `features` tests, then run locally against local repo (instructions in base README)
  - with something like `go run . run --lang go --version ../sdk-go --no-history-check worker_shutdown/poll_complete_on_shutdown`
3. If regressions found, have opportunity to merge a fix in before cutting a release
4. Cut release
5. Merge `features` tests
6. All subsequent PRs will have new and legacy test coverage.

## Why?

More test coverage for legacy behavior

## Checklist

1. Closes #835

7. How was this tested:
Added a bunch of tests

8. Any docs updates needed?


Sushisource (approved):
Overall seems perfectly sensible to me

My only nit is `config.json` doesn't need to have a `.` prefix - no reason to hide it.

PR #840, opened 2026-05-28
Add Async Workflow Nexus Test

Quinn-With-Two-Ns:
## What was changed

Add Async Workflow Nexus Test

## Why?

Add Coverage for async nexus operations

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/nexus/workflow_run_success/feature.go
+++ b/features/nexus/workflow_run_success/feature.go
@@ -1,0 +1,137 @@
+package workflow_run_success
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"github.com/temporalio/features/harness/go/harness"
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	historypb "go.temporal.io/api/history/v1"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/temporalnexus"
+	"go.temporal.io/sdk/workflow"
+)
+
+const ServiceName = "test-service"
+
+func HandlerWorkflow(ctx workflow.Context, name string) (string, error) {
+	return "Hello, " + name + "!", nil
+}
+
+var AsyncWorkflowOperation = temporalnexus.NewWorkflowRunOperation(
+	"AsyncWorkflowOperation",
+	HandlerWorkflow,
+	func(ctx context.Context, input string, opts nexus.StartOperationOptions) (client.StartWorkflowOptions, error) {
+		// Use the request ID so retried start requests resolve to the same workflow.
+		return client.StartWorkflowOptions{ID: "nexus-async-workflow-" + opts.RequestID}, nil
+	},
+)
+
+var Service = func() *nexus.Service {
+	s := nexus.NewService(ServiceName)
+	s.MustRegister(AsyncWorkflowOperation)
+	return s
+}()
+
+func CallerWorkflow(ctx workflow.Context, endpoint string) (string, error) {
+	nc := workflow.NewNexusClient(endpoint, ServiceName)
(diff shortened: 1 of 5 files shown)


features/nexus/workflow_run_success/feature.go
```diff
@@ -0,0 +1,137 @@
...
+		if completed, err := findCallerEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_COMPLETED); err != nil {
+			return err
+		} else if completed == nil {
+			return fmt.Errorf("did not find NexusOperationCompleted event in history")
+		}
+
+		// The caller's NexusOperationStarted event must link to the handler workflow's
+		// WorkflowExecutionStarted event.
+		handlerLink := findWorkflowEventLink(started.GetLinks(), enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED)
+		if handlerLink == nil {
```
KeithB-Temporal, line 91: We don't want to validate anything else about the link except that it exists?

features/nexus/workflow_run_success/feature.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("handler callback link references %s/%s, expected caller %s/%s",
+				callerLink.GetWorkflowId(), callerLink.GetRunId(), run.GetID(), run.GetRunID())
+		}
+		return nil
+	},
+}
+
+// findWorkflowEventLink returns the first WorkflowEvent-variant link whose event reference matches
+// the given event type, or nil if none match.
+func findWorkflowEventLink(links []*commonpb.Link, eventType enumspb.EventType) *commonpb.Link_WorkflowEvent {
```
KeithB-Temporal, line 150: Not sure if this is a valid concern (because not sure if we ever run tests that could use this workflow events space, but if it always just returns the first link this could make this test not parallelizable (if any other test caused links to get into workflow events with that event type.)  Ignore my comment if no other tests could side effect this.

features/nexus/workflow_run_success/feature.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public void checkHistory(Runner runner, Run run) throws Exception {
+      // Async (workflow-run) Nexus operations should transition Scheduled -> Started -> Completed.
+      var events = runner.getWorkflowHistory(run).getEventsList();
+      var scheduled = findEvent(events, e -> e.hasNexusOperationScheduledEventAttributes());
+      assertNotNull(scheduled, "expected NexusOperationScheduled event in history");
+      var started = findEvent(events, e -> e.hasNexusOperationStartedEventAttributes());
+      assertNotNull(started, "expected NexusOperationStarted event in history");
+      assertTrue(
+          events.stream().anyMatch(e -> e.hasNexusOperationCompletedEventAttributes()),
```
KeithB-Temporal: Why do we findEvent and assertNotNull for scheduled and started, but for OperationCompleted we check the stream for matches and assertTrue that we found one?

KeithB-Temporal (approved):
Overall changes look good. Some minor questions about parrallelizability and validation.

features/nexus/workflow_run_success/feature.java
Quinn-With-Two-Ns: We don't need to inspect the `NexusOperationCompletedEventAttributes`, I can change it for consistency the code isn't much longer

features/nexus/workflow_run_success/feature.go
Quinn-With-Two-Ns: No, every test would run in isolation so no risk of one test effecting another

features/nexus/workflow_run_success/feature.go
Quinn-With-Two-Ns: Initially I didn't bother since most of the content of the link is random like the workflow ID or run ID, but we can at least easily assert the namespace and part of the workflow ID looks right

PR #871, opened 2026-08-19
feat(nexus): add async, cancellation, failure and parallel features

xepozz:
Four Nexus scenarios implemented in Go and Java: an async workflow-run operation, cancelling a running async operation, a sync operation that fails with an application error, and three parallel sync operations.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


features/nexus/async_cancellation/feature.java
```diff
@@ -0,0 +1,136 @@
...
+      var scope =
+          Workflow.newCancellationScope(
+              () -> {
+                var handle = Workflow.startNexusOperation(stub::blockingOperation, "world");
+                handle.getExecution().get();
+                handle
+                    .getResult()
+                    .handle(
+                        (value, failure) -> {
+                          resultHolder[0] = failure == null ? "completed" : "cancelled";
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: This is a good flag, in the Go test we do check for this already

features/nexus/async_cancellation/feature.java
```diff
@@ -0,0 +1,136 @@
...
+          NexusServiceOptions.newBuilder()
+              .setEndpoint(endpoint)
+              .setOperationOptions(
+                  NexusOperationOptions.newBuilder()
+                      .setScheduleToCloseTimeout(Duration.ofMinutes(1))
+                      .build())
+              .build();
+      TestService stub = Workflow.newNexusServiceStub(TestService.class, serviceOptions);
+
+      var resultHolder = new String[1];
```
Quinn-With-Two-Ns:
```suggestion
      var resultHolder = Workflow.newPromise();
```

features/nexus/sync_operation_error/feature.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @ServiceImpl(service = TestService.class)
+  class TestServiceImpl {
+    @OperationImpl
+    public OperationHandler<String, String> failingOperation() {
+      return OperationHandler.sync(
+          (context, details, name) -> {
+            throw OperationException.failure(
+                ApplicationFailure.newNonRetryableFailure("deliberate failure", "TestError"));
```
Quinn-With-Two-Ns: In the Go test  `temporal.NewApplicationError(ErrorMessage, ErrorType)` this will return a retry able application error

Quinn-With-Two-Ns: The two tests should be consistent

features/nexus/sync_operation_error/README.md
```diff
@@ -0,0 +1,13 @@
+# Nexus sync operation fails
+
+A workflow invokes a synchronous Nexus operation that raises an application failure and
```
Quinn-With-Two-Ns: The Nexus operations you wrote do not "raises an application failure", they raise a `OperationError`. The `OperationError` happens to contain a application failure as a cause but from the SDKs perspective that is not relevant. What you have tested is fine, but I would correct the spec. 

I'll note having a test where the Nexus operation DOES raise a application failure would also be a good test, but it doesn't have to be in this PR

features/nexus/parallel_sync_operations/README.md
```diff
@@ -0,0 +1,12 @@
+# Nexus operations run in parallel
+
+A workflow starts three synchronous Nexus operations in a single workflow task and awaits all
```
Quinn-With-Two-Ns, line 3: Can you please change the folder name to clarify this is parallel sync operations.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
PHP test reliably failing on this PR

PR #879, opened 2026-09-01
External storage: Test oversized query result storage

yuandrew:
## What was changed

Verify supported SDKs offload oversized query results before applying the server payload limit.

## Why?
Go had a bug, might as well write cross-SDK tests to verify this is Go only bug 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


package.json
```diff
@@ -8,17 +8,11 @@
     "start": "ts-node -r tsconfig-paths/register harness/ts/main.ts"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.0.0",
-    "@google-cloud/storage": "^7.0.0",
     "@grpc/grpc-js": "^1.12.4",
     "@protobuf-ts/protoc": "^2.8.1",
     "@temporalio/activity": "^1.21.1",
```
jmaeagle99, line 13: All of these should already be updated to 1.21.1 in main.

features/query/oversized_result_external_storage/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "go": {
+    "minVersion": "1.48.0"
```
jmaeagle99, line 3: This only works for Go and not for any other language. See https://github.com/temporalio/features/blob/main/harness/go/cmd/run.go#L103. Probably either need to add harness support or do runtime skips. I'm okay with the latter on this for now. I can follow up with general config support.

jmaeagle99, line 3:
```suggestion
    "minVersion": "v1.48.0"
```

features/query/oversized_result_external_storage/feature.go
```diff
@@ -0,0 +1,142 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	queryName    = "oversized-result"
+	finishSignal = "finish"
+	driverName   = "query-result-memory"
+	// Exceed the server limit so the SDK must offload the result.
+	resultSize       = 3 * 1024 * 1024
+	storageThreshold = 1024
```
jmaeagle99, line 23: The default threshold is 256 KiB, which is much lower than the result size for each of these tests. Probably do not need to specify the threshold explicitly.

features/query/oversized_result_external_storage/feature.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+	if result != strings.Repeat("a", resultSize) {
+		return fmt.Errorf("unexpected query result")
+	}
+	if storage.storeCalls() == 0 || storage.retrieveCalls() == 0 {
+		return fmt.Errorf("query result did not use external storage")
+	}
+
+	if err := r.Client.SignalWorkflow(ctx, run.GetID(), run.GetRunID(), finishSignal, nil); err != nil {
```
jmaeagle99, line 68: Should this be done in a `finally` to make sure we let the workflow close if an assertion is failed earlier. Not in a `defer` because `CheckResultDefault` waits for the run result and would deadlock.

features/query/oversized_result_external_storage/feature.py
```diff
@@ -0,0 +1,103 @@
...
+        for claim in claims:
+            payload = Payload()
+            payload.ParseFromString(self.payloads[claim.claim_data["key"]])
+            payloads.append(payload)
+        return payloads
+
+
+driver = MemoryDriver()
+data_converter = dataclasses.replace(
+    temporalio.converter.default(),
```
jmaeagle99, line 63: This is deprecated. Use `DataConverter.default`.

features/query/oversized_result_external_storage/feature.go
```diff
@@ -0,0 +1,142 @@
...
+		key := claim.ClaimData["key"]
+		payload, ok := d.payloads[key]
+		if !ok {
+			return nil, fmt.Errorf("payload %q not found", key)
+		}
+		payloads[i] = proto.Clone(payload).(*commonpb.Payload)
+	}
+	return payloads, nil
+}
+
```
jmaeagle99, line 131: Maybe make a `func (d *memoryDriver) calls() (int, int)` func so we get both values under the same lock to avoid tearing.

features/query/oversized_result_external_storage/feature.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	var result string
+	if err := value.Get(&result); err != nil {
+		return err
+	}
+	if result != strings.Repeat("a", resultSize) {
+		return fmt.Errorf("unexpected query result")
+	}
+	if storage.storeCalls() == 0 || storage.retrieveCalls() == 0 {
```
jmaeagle99, line 64: `storage` is package scoped so if this is run multiple times due to variants or any future fanout or sharing across tests, these assertions won't hold. Probably need to snapshot the storage counts before executing the query, snapshot again after, and make sure the difference is what we expect (likely should be 1 for store and retrieve).

features/query/oversized_result_external_storage/feature.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+async def check_result(_: Runner, handle: WorkflowHandle) -> None:
+    result = await handle.query(Workflow.oversized_result)
+    assert result == "a" * RESULT_SIZE
+    assert driver.stores > 0
+    assert driver.retrieves > 0
+
+    await handle.signal(Workflow.finish)
+    await handle.result()
```
jmaeagle99, line 96: Doesn't the harness do this already?

features/query/oversized_result_external_storage/feature.go
```diff
@@ -0,0 +1,142 @@
...
+func newMemoryDriver() *memoryDriver {
+	return &memoryDriver{payloads: make(map[string]*commonpb.Payload)}
+}
+
+func (*memoryDriver) Name() string {
+	return driverName
+}
+
+func (*memoryDriver) Type() string {
+	return driverName
```
jmaeagle99, line 91: Should we use a unique test driver name for features, in case these are run against cloud? This is going to be reported on worker heartbeat.

features/query/oversized_result_external_storage/README.md
```diff
@@ -0,0 +1,9 @@
+# Oversized query result with external storage
+
+An oversized query result succeeds when external storage is configured. The
+SDK must offload the result before applying the server payload-size limit.
+
+The query returns 3 MiB, above the server's default 2 MiB error limit. The
+checker verifies the value and confirms that storage and retrieval occurred.
+
+Go, Python, and TypeScript currently expose functional external storage.
```
jmaeagle99, line 9: Probably can remove this as that's demonstrated by what's implemented as tests in here.

jmaeagle99 (changes requested):

PR #880, opened 2026-09-02
Add Transfer Types Spec

VegetarianOrc:
## What was changed

- Add generated spec based on Python, .NET, and Java prototype behavior for collaborative review

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/features/data_converter/transfer_types/spec.md
+++ b/features/data_converter/transfer_types/spec.md
@@ -73,7 +73,7 @@ else:
 
 ## Behaviors
 
-### Conversion wraps configured payload converters
+### Conversion Flow
 
 #### Encoding
 
@@ -144,13 +144,15 @@ sequenceDiagram
 3. Ask the configured payload converter to decode the payload as the transfer type.
 4. Invoke the transfer converter to reconstruct the requested model value.
 
-### Top-level values only
+### SDK Behavior
+
+#### Top-level values only
 
 Transfer conversion only applies to top-level values and does not recursively inspect fields for transfer types.
 
 Nested serialization is the responsibility of the configured payload converter.
 
-### At most one transfer step
+#### At most one transfer step
 
 Transfer type converters are applied at most once. For example, given:
 
@@ -162,7 +164,7 @@ When encoding type `A`, the SDK does not inspect the output type `B` for a trans
 
 Similarly when decoding, the SDK will convert using only the transfer type encoder registered on the requested model type.
 
-### TransferTypeConverter Selection
+#### TransferTypeConverter Selection
 
 Converting to the transfer type should use the type hint for the target execution rather than the value's runtime type. For example:
 
@@ -188,11 +190,11 @@ A transfer converter declaration applies only to the exact model type on which i
 
 A subclass or other derived type can declare its own converter independently of its base type.
 
-### TransferType Nullability Converter
+#### Transfer Type Nullability
 
 TransferTypeConverters should specify a non-null transfer type.
 
-### Failure conversion
+#### Failure conversion
 
 Failure conversion should use the transfer converter before
 applying the configured payload converter to failure details.
--- a/features/data_converter/transfer_types/spec.md
+++ b/features/data_converter/transfer_types/spec.md
@@ -4,7 +4,7 @@ Last updated: 2026-09-02
 
 ## Motivation
 
-Transfer Type Conversion provides a way to translate to and from a Payload serialization friendly format, e.g. a protobuf object, to a more ergonomic or language idiomatic type.
+Transfer Type Conversion provides a way to translate between a Payload serialization friendly format, e.g. a protobuf object, and a more ergonomic or language-idiomatic type.
 TransferTypeConverters are added to the API facing type and specify the target "transfer type" which is then used by the configured Payload Converter serialize to/from a Payload.
 
 For example, given the protobuf definition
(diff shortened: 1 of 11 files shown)


features/transfer_types/spec.md
```diff
@@ -0,0 +1,172 @@
+# Transfer Type Converter Behavioral Specification
+
+Status: Draft
+
+Last updated: 2026-09-02
+
+## Purpose
+
+This document defines the portable behavior of Temporal transfer type converters. It specifies observable conversion semantics without prescribing SDK-specific APIs, generic type mechanisms, or converter construction strategies.
```
dplyukhin:
```suggestion
## Motivation

Sometimes you have a custom datatype that should be serializable as a payload, but you can't modify the configured data converter to support your type. Transfer types let you work around that by picking a "transfer type" that's already serializable, and associating your datatype with a "transfer type converter" that converts values of your custom type to values of the transfer type and back.
```

features/transfer_types/spec.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Terminology
+
+- **Model type:** The user-facing type associated with a transfer type converter.
+- **Transfer type:** The serializer-facing type produced from a model value.
+- **Transfer value:** A value of the transfer type.
+- **Payload conversion:** Conversion between an in-memory value and a Temporal `Payload`.
+- **Payload codec:** A byte-to-byte transformation applied to a payload, such as encryption or compression.
+- **Failure details:** Typed payloads stored in application failures, cancellations, timeout heartbeat details, reset failures, or nested failure causes.
```
dplyukhin:
```suggestion
- **Model type:** The user-facing type associated with a transfer type converter.
- **Transfer type:** The serializable type produced from a model value.
- **Transfer value:** A value of the transfer type.
```

features/transfer_types/spec.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Normative behavior
```
dplyukhin:
```suggestion
## Behavior
```

dplyukhin (changes requested):

features/data_converter/transfer_types/spec.md
```diff
@@ -0,0 +1,198 @@
+# Transfer Type Converter Behavioral Specification
+
+Last updated: 2026-09-02
+
+## Motivation
+
+Transfer Type Conversion provides a way to translate to and from a Payload serialization friendly format, e.g. a protobuf object, to a more ergonomic or language idiomatic type.
```
dplyukhin:
```suggestion
Transfer Type Conversion provides a way to translate between a Payload serialization friendly format, e.g. a protobuf object, and a more ergonomic or language-idiomatic type.
```

features/data_converter/transfer_types/spec.md
```diff
@@ -0,0 +1,198 @@
...
+A subclass or other derived type can declare its own converter independently of its base type.
+
+### TransferType Nullability Converter
+
+TransferTypeConverters should specify a non-null transfer type.
+
+### Failure conversion
+
+Failure conversion should use the transfer converter aware payload converter before
+applying the configured payload converter to failure details.
```
dplyukhin: I thought we wanted this?
```suggestion
Failure conversion should use the transfer converter before
applying the configured payload converter to failure details.
```

features/data_converter/transfer_types/spec.md
```diff
@@ -0,0 +1,211 @@
...
+
+Similarly when decoding, the SDK will convert using only the transfer type encoder registered on the requested model type.
+
+#### Transfer Type Nullability
+
+TransferTypeConverters should specify a non-null transfer type.
+
+#### Failure conversion
+
+Failure conversion should NOT use the transfer converter.
```
dplyukhin: I'm forgetting, didn't we say memos shouldn't use the transfer converter?

```suggestion
#### Payloads that don't use the transfer converter

Failure conversion and memos should NOT use the transfer converter.
```

features/data_converter/transfer_types/spec.md
```diff
@@ -0,0 +1,211 @@
...
+    print("No workflow execution provided")
+```
+
+## Terminology
+
+- **Model type:** The user-facing type associated with a transfer type converter.
+- **Model value:** A value of the model type.
+- **Transfer type:** The serializer-facing type produced from a model value.
+- **Transfer value:** A value of the transfer type.
+- **Transfer type converter:** The conversion logic to map between the model type and the transfer type.
```
dplyukhin: Tim's short name for it is occasionally nice to have
```suggestion
- **Transfer type converter:** The conversion logic to map between the model type and the transfer type.

We occasionally refer to "transfer type conversion" as "transfer conversion" for short.
```

features/data_converter/transfer_types/spec.md
VegetarianOrc: I think it's reasonable to exclude them, we can always add it if there winds up being a real use.

features/data_converter/transfer_types/spec.md
dplyukhin: Would you consider the nickname "transfer converter" in place of "transfer type converter"?

VegetarianOrc: I don't have a strong opinion there and find myself using them interchangeably. Happy to switch unless @tconley1428 feels we shouldn't

dplyukhin: Let's keep both terms then :) I like having a short name, but we've also used the long name all over the place so it's too late to change anyway

dplyukhin (approved):

PR #892, opened 2026-09-06
Bump protobufjs pnpm override 7.5.1 → ^8.7.1

rkannan82:
## What
Bump `protobufjs` from `7.5.1` to `^8.7.1` in both the root `package.json` and the pnpm override in `sdkbuild/typescript.go`.

## Why
`@temporalio/common@1.23.0` requires `protobufjs@^8.7.1`. The 7.5.1 pin (added as a temporary workaround in #625, underlying issue since fixed in temporalio/sdk-typescript#1717) blocks 8.x resolution, breaking all TS kitchensink tests in omes with `Cannot find module 'protobufjs/ext/protojson'`.

## How did you test it?
Confirmed `@temporalio/common@1.23.0` declares `protobufjs@^8.7.1` via `npm view`. Confirmed `ext/protojson` was added in protobufjs 8.x. Will verify omes CI passes after this merges and omes bumps its features dep.


stephanos:
I also need this fix 👀

THardy98 (approved):

PR #896, opened 2026-09-09
docs: add System Nexus design

tconley1428:
Adds a cross-SDK System Nexus design note covering envelope serialization, nested payload handling, serialization context, interception, and TypeScript-specific considerations.


features/system_nexus/README.md
```diff
@@ -0,0 +1,39 @@
+# System Nexus
+
+System Nexus operations are normal nexus operations which happen to target the `__temporal_system` endpoint. They are handled by the server rather than a user's own defined nexus handler.
+
+## Serialization
+
+Because the operations need to be read by the server, the nexus operation's input, which we'll call the "system nexus envelope" needs to be handled uniquely. It can't be serialized with the user's data converter (payload converter, codec, or external storage) because the server would then be unable to read it. Instead, it needs to reach the server specifically encoded with binary protobuf. It should additionally have `__temporal_system_payload` metadata for use in later recognition.
```
dplyukhin, line 7:
```suggestion
Because the operations need to be read by the server, the nexus operation's input, which we'll call the "system nexus envelope" needs to be handled uniquely. It can't be serialized with the user's data converter (payload converter, codec, or external storage) because the server would then be unable to read it. Instead, it needs to reach the server specifically encoded with binary protobuf. It should additionally have `__temporal_system_payload = "true"` metadata for reasons that will become clear in the next section.
```

dplyukhin: What do you mean by "later recognition"?

dplyukhin:
Super helpful! Thanks for writing this. Better get some more eyes on it besides just me though

VegetarianOrc (approved):

